PR #14, opened 2017-04-18
ZCS-1092 Inactive filters should not be displayed in active list, if activation fails

silentsakky:
- if user tries to activate filter which contains operation to move mails to a deleted folder then show error to user and reload the list to rollback ui changes
- use pre existing error message instead of creating new one

### DIFF
--- a/WebRoot/js/zimbraMail/prefs/model/ZmFilterRules.js
+++ b/WebRoot/js/zimbraMail/prefs/model/ZmFilterRules.js
@@ -376,20 +376,36 @@ function(index, notify, callback, result) {
 ZmFilterRules.prototype._handleErrorSaveRules =
 function(ex) {
 	if (ex.code == ZmCsfeException.SVC_PARSE_ERROR ||
-		ex.code == ZmCsfeException.SVC_INVALID_REQUEST)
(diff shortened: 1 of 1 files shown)


WebRoot/js/zimbraMail/prefs/model/ZmFilterRules.js
```diff
@@ -376,11 +376,17 @@ function(index, notify, callback, result) {
 ZmFilterRules.prototype._handleErrorSaveRules =
 function(ex) {
 	if (ex.code == ZmCsfeException.SVC_PARSE_ERROR ||
-		ex.code == ZmCsfeException.SVC_INVALID_REQUEST)
+		ex.code == ZmCsfeException.SVC_INVALID_REQUEST ||
+		ex.code == ZmCsfeException.MAIL_NO_SUCH_FOLDER)
```
rohan-ambasta: You may also get similar exception if a tag does not exist. Are you handling that?

silentsakky: Thanks @rohan-ambasta, I have updated code to handle that scenario as well

PR #18, opened 2017-04-21
Fix issue of address bubble and tag bubble in mail view reading pane.

kalpeshhpatel:
Fix issue of address bubble and tag bubble in mail view reading pane.

Before:
screen shot 2017-04-21 at 11 04 10 am

After fix:
screen shot 2017-04-21 at 11 25 59 am


tarangkhandelwal:
I see the space between the two rows : From & To reduced with the fix. Is this expected?

kalpeshhpatel:
@tarangkhandelwal , yes, it is expected as per mockup. Initially it was like that only.

Thanks

PR #26, opened 2017-05-02
cleaned up things

vschs007:
> indentation adjustments
> added space in top


silentsakky:
This is not playground for kids :)

vschs007:
Yeah, i never thought of it :P

PR #29, opened 2017-05-05
ZCS-912 Universal UI: Attach briefcase item modal from compose view

tarangkhandelwal:
Changeset:

1. Moving svg coloring styles from zm.css to skin.css to keep all similar styles at one place.
2. ZmBriefcaseTabView.js: Updating the dialog with from 500 to 600.
3. ZmBriefcaseIconView: Getting icon using AjxImg utility.
4. zm.css & skin.css : Adding styles for briefcase row item as per UX mockups.

### DIFF
--- a/WebRoot/css/common.css
+++ b/WebRoot/css/common.css
@@ -589,6 +589,7 @@ LI.RowDouble, LI.Row {list-style: none;}
 /* .Row-matched TD			 	{ 	@ListItemText-matched@				} */
 
 .Row-focused TD 				{	@ListItemText-focused@			}
+.Row-selected.Row-focused TD 				{	@ListItemText-selected@			}
 
 .Row-selected-right TD			{	@ListItemText-actioned@			}
 
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -4176,6 +4176,7 @@ HTML>BODY .appt-selected .appt_allday_body
 
 .ZmAttachDialog-inline {
 	height:20px;
+	margin-top:12px;
 }
 
 .ZmAttachDialog-removeLink {
@@ -5236,9 +5237,27 @@ If you feel compelled to have a background color, please do so on a per skin lev
 }
 
 .ZmAttachDialog-container .ZmBriefcaseItemSmall {
-	border-style:solid; 
-	border-width:1px 0px;
-	border-color:@lighten(AppC,60)@
+	@AttachBriefcaseItem@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemSmall.Row-selected {
+	@AttachBriefcaseItemSelected@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:nth-child(2),
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:nth-child(3) {
+	@AttachBriefcaseItemChild@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:first-child {
+	@AttachBriefcaseItemFirstChild@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemIcon .svg-icon {
+	@AttachBriefcaseItemIcon@
+}
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td {
+	@TreeItemHeight@;
 }
 
 .ZmBriefcaseItemSelected .ZmThumbnailItem {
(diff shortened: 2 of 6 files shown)


WebRoot/css/zm.css
```diff
@@ -5242,9 +5242,24 @@ If you feel compelled to have a background color, please do so on a per skin lev
...
+.ZmAttachDialog-container .ZmBriefcaseItemSmall.Row-selected {
+	@AttachBriefcaseItemSelected@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:nth-child(2),
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:nth-child(3) {
+	@AttachBriefcaseItemChild@
+}
+
+.ZmAttachDialog-container .ZmBriefcaseItemSmall td:first-child {
```
silentsakky: accessing elements using first/last child would be risky, as skin developer can change positioning of elements using template and these styles will be applied to wrong elements then

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1773,7 +1778,7 @@ CheckboxTableRightText      = text-align:left; margin-left:@CheckBoxTableCellSpa
 MailAttachmentsTable        = @FullWidth@
 MailAttachmentItem          = @FullWidth@ @AppBg@ margin-bottom:2px; @roundCorners(3px)@ overflow:hidden;
 MailAttachmentItemActive    = background-color: @darken(AppC,13)@;
-MailAttachmentIcon          = @BigIconSize@
+MailAttachmentIcon          = @BigIconSize@; @roundCorners(0px)@ !important;
```
silentsakky: can we remove important

WebRoot/css/zm.css
tarangkhandelwal: In this case it is not expected to have specific elements on first or second position. We are just applying the padding for breathing space which is independent of which element is first . So we should be good with this :)

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: We required this to override the generic styling for roundCorners coming from "FileIconCommonStyle" . We need no roundCorners only in mail reading view. 

@kalpeshhpatel  can also help on this maybe.

WebRoot/css/zm.css
kalpeshhpatel: I feel there is no risk to use `:first-child` and `:last-child` pseudo selectors for such things, unless we target for major style declarations.

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: Agreed with @tarangkhandelwal as we need to override default rule for this. Also this rule doesn't have any side-effects, as it targets only single element.

WebRoot/css/common.css
```diff
@@ -589,6 +589,7 @@ LI.RowDouble, LI.Row {list-style: none;}
 /* .Row-matched TD			 	{ 	@ListItemText-matched@				} */
 
 .Row-focused TD 				{	@ListItemText-focused@			}
+.Row-selected.Row-focused TD 	{	@ListItemText-selected@			}
```
kalpeshhpatel: Can you please correct indentation?

WebRoot/css/zm.css
```diff
@@ -5242,9 +5242,24 @@ If you feel compelled to have a background color, please do so on a per skin lev
 }
 
 .ZmAttachDialog-container .ZmBriefcaseItemSmall {
-	border-style:solid; 
-	border-width:1px 0px;
-	border-color:@lighten(AppC,60)@
+	@AttachBriefcaseItem@
```
kalpeshhpatel: 👍

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1217,6 +1217,11 @@ ComposeViewHtmlEditor       = padding:@ComposeViewOutsidePadding@; background-co
 ComposeViewPlainTextEditor  = @FontSize-big@ @FontFamily-fixed@ @FullWidth@ @AppBorder@ overflow:auto; padding:8px;
 ComposeViewHeaderEmptyRight = padding-right: 18px;
 
+AttachBriefcaseItem         = border-style:solid; border-width:0 0 1px 0; border-color:transparent;
+AttachBriefcaseItemSelected = border-bottom-color:@PanelBorderColor@;
+AttachBriefcaseItemChild    = padding-left:@MailViewInnerSpacing@;
```
kalpeshhpatel: If we don't need padding for other sides i.e 0px from all side expect left, can we have style like `padding:0 0 0 @MailViewInnerSpacing@`, to avoid the possibility of such styles applied accidentally by some other selectors. 

This is just suggestion, feel free to keep as it is, if you feels that won't be case and it's okay.

Thanks

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1217,6 +1217,11 @@ ComposeViewHtmlEditor       = padding:@ComposeViewOutsidePadding@; background-co
...
+AttachBriefcaseItemFirstChild = padding-left:@SkinWrapperPadding@;
```
kalpeshhpatel: Same suggestions as above. Again feel free to keep as it is, if you feels that okay.

Thanks

WebRoot/skins/_base/base4/skin.properties
silentsakky: Can we increase specificity of MailAttachmentIcon selector instead of using important?

tarangkhandelwal: Sure .Thanks

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: Yes this would be better.Thanks

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: Yes this would be better.Thanks

WebRoot/css/common.css
tarangkhandelwal: I can add more tabs to match the above line. Though its not very good with tabs in here :)

WebRoot/css/zm.css
```diff
@@ -6529,7 +6487,7 @@ a.ZmLinkDisabled {
 }
 
 .ZmMailAttachmentIcon,
-.ZmMailAttachmentIcon .svg-icon {
+.ZmMailAttachmentIcon svg.svg-icon {
```
silentsakky: minor, should this go to skin.css as all svg icon related properties are moved there?

kalpeshhpatel: I think we are moving only direct `.svg-icon.*` kind of rules, which will be common for all SVG's.
However, some SVG in some containers may need some customization, so I think it's better if keep such rules together for logical grouping of rules.

silentsakky: ok

tarangkhandelwal:
Squashed the commits and merging it .

PR #31, opened 2017-05-08
ZCS-1304

SatyamKumarOnline:
Cross-Site Scripting vulnerability in Zimbra exploitable by sending a specially crafted email

Desc -
HTML-encode for the description and location field.
Also, adding content to gitignore file.

Changeset -

ZmMailMsgView.js
ZmInvite.js

Testing notes  -
Ran the below test cases successfully against the fix.

Case 1:
1. Send a mail with subject 'My message' from user1 to user2.
2. Inject the attached mime_XSS1 (script is present in location field) to user2
3. Login as user2 and select the conversation.
script present in location field should not be executed.

Case 2:
1. Change the view from conversation to message
2. Load Inbox and select the injected message in case 1.
script present in location field should not be executed.

case 3.
1. Send a mail with subject 'My message3' from user1 to user2.
2. Inject the attached mime_XSS3 (script is present in description field) to user2
3. Login as user2 and select the conversation.
script present in description field should not be executed.

Case 4:
1. Change the view from conversation to message
2. Load Inbox and select the injected message case 3.
script present in description field should not be executed.

Case 5:
Repeat case 1 and case 2 using mime mime_XSS4 with subject 'My message4' to verify XSS on organizer field

Case 6:
Repeat case 1 and case 2 using mime mime_XSS5 with a subject 'My message5' to verify XSS on attendee field.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 build/
+.project
+/.history
+/.idea
--- a/WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
+++ b/WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
@@ -1831,6 +1831,7 @@ ZmMailMsgView.prototype._renderMessageBody1 = function(params, part) {
         // calendar part in ICS format
         else if (ct === ZmMimeTable.TEXT_CAL) {
             content = ZmMailMsg.getTextFromCalendarPart(bp);
+            content = AjxStringUtil.htmlEncode(content);
         }
 
         // HTML
--- a/WebRoot/js/zimbraMail/share/model/ZmInvite.js
+++ b/WebRoot/js/zimbraMail/share/model/ZmInvite.js
@@ -989,7 +989,7 @@ function(isHtml) {
 		buf[i++] = "<p>\n<table border='0'>\n";
 	}
 
-	var orgName = this.getOrganizerName();
+	var orgName = AjxStringUtil.htmlEncode(this.getOrganizerName());
 	if (orgName) {
 		params = [ZmMsg.organizerLabel, orgName, ""];
 		buf[i++] = formatter.format(params);
@@ -1003,7 +1003,7 @@ function(isHtml) {
 		buf[i++] = "\n";
 	}
 
-	var locationSummary = this.getLocation();
+	var locationSummary = AjxStringUtil.htmlEncode(this.getLocation());
 	if (locationSummary) {
 		params = [ZmMsg.locationLabel, locationSummary, ""];
 		buf[i++] = formatter.format(params);


.gitignore
```diff
@@ -1 +1,11 @@
 build/
+.project
+/WebRoot/js/zimbra/
```
tarangkhandelwal: I would suggest not to add js folders that are currently copied from zm-ajax into .gitignore. 

Also, ant clean removes these if required.

WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
```diff
@@ -1831,6 +1831,7 @@ ZmMailMsgView.prototype._renderMessageBody1 = function(params, part) {
         // calendar part in ICS format
         else if (ct === ZmMimeTable.TEXT_CAL) {
             content = ZmMailMsg.getTextFromCalendarPart(bp);
+            content = AjxStringUtil.htmlEncode(content);
```
tarangkhandelwal: shouldn't the htmlEncode function be added into ZmMailMsg.getTextFromCalendarPart function to return encoded string ? There could be calls to this function from multiple places  which would get solved with that.

plobbes: Short answer, probably not.  Longer answer, maybe but it depends on the documented behavior and actual use cases.  In general, it's ok for the data to be returned in a "pristine" form, but the caller needs to be sure that the data gets encoded properly for the context where that data is used.

Given the name getText... I suspect that we shouldn't change that function.

silentsakky: I saw one change mentioned in ticket related to AjxStringUtil._getOriginalHtmlContent function, is that change not required here?

SatyamKumarOnline: ZmMailMsg.getTextFromCalendarPart  is being called from here only.
More ever, getTextFromCalendarPart  gives me the text content out of a calendar part, so it's good to not use it there.
More ever, the context here is while rendering of message, when the malicious script of 
Description field is getting executed inside hidden Iframe. And, not in other cases.
So, better to take care of encoding here, itself.

Thanks.

SatyamKumarOnline: Change in AjxStringUtil._getOriginalHtmlContent  is not required, here.
For XSS issues, we usually encode the corresponding malicious fields.
AjxStringUtil._getOriginalHtmlContent  is subsequently called in further function calls.

.gitignore
SatyamKumarOnline: I have added js folders that are currently copied from zm-ajax into .gitignore as lot of untracked files were showing up while locally committing the changes .
Running ant clean explicitly seems overdue to me each time you plan to do  commit.
The gitignore file that is in the universal UI branch has same content.
https://github.com/Zimbra/zm-web-client/blob/feature/universal-ui/.gitignore

However, as discussed offline, as there is again some restructuring of repo that is going to happen for ajax client near, removed the copied zm-ajax  folders from gitignore.

.gitignore
```diff
@@ -1 +1,5 @@
 build/
+.project
+/.gitignore
```
tarangkhandelwal: we should also remove /.gitignore . Not sure of a case when this should be ignored.

WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
tarangkhandelwal: If that's only been called once, we can go ahead with this version also considering the point about keeping data pristine from @plobbes .

silentsakky:
found issue where xss was possible using organizer/attendee fields

.gitignore
SatyamKumarOnline: Done

WebRoot/js/zimbraMail/share/model/ZmInvite.js
```diff
@@ -295,8 +295,8 @@ function(compNum) {
 ZmInvite.prototype.getOrganizerName = 
 function(compNum) {
 	var cn = compNum || 0;
-	return (this.components[cn] && this.components[cn].or)
-		? (this.components[cn].or.d || this.components[cn].or.url) : null;
+	return AjxStringUtil.stripTags((this.components[cn] && this.components[cn].or)
```
tarangkhandelwal: Can we have the value in a variable then call the stripTags on that?  Doing so much in a single line (condition check, value assignment & passing through utility function) makes it difficult to understand for anyone going through or debugging.

tarangkhandelwal:
@SatyamKumarOnline  : I see we using different utility functions to fix the issue for different functions _AjxStringUtil.htmlEncode_ & _AjxStringUtil.stripTags_ . Is there a reason for this?

WebRoot/js/zimbraMail/share/model/ZmInvite.js
SatyamKumarOnline: Done

WebRoot/js/zimbraMail/share/model/ZmInvite.js
```diff
@@ -853,7 +855,7 @@ function(compNum) {
 ZmInvite.prototype.getLocation =
 function(compNum) {
 	var cn = compNum || 0;
-	return this.components[cn] ? this.components[cn].loc : null;
+	return this.components[cn] ? AjxStringUtil.stripTags(this.components[cn].loc) : null;
```
tarangkhandelwal: As discussed, this should be done at the UI level and AjxStringUtil.htmlEncode should be used as in this case email is getting lost when its wrapped on ui with  <> .
@SatyamKumarOnline

SatyamKumarOnline: Done.

PR #33, opened 2017-05-10
ZCS-1298 must login every time browser restarts

silentsakky:
- in case of two factor auth is enabled and user has selected remember me option on login page
- so we are showing additional screen for getting totp, and after totp is validated we need to rewrite ZM_AUTH_TOKEN with proper expires headers as per value of remember me option

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -459,6 +459,11 @@ if (application.getInitParameter("offlineMode") != null) {
 					<c:otherwise>
 								<form method="post" name="loginForm" action="${formActionUrl}" accept-charset="UTF-8">
 								<input type="hidden" name="loginOp" value="login"/>
+
+								<c:if test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">
+									<!-- if user has selected remember me in login page and we are showing totp screen to user, then we need to maintain value of that flag as after successfull two factor authentication we will have to rewrite ZM_AUTH_TOKEN with correct expires headers -->
+									<input type="hidden" name="zrememberme" value="${param.zrememberme}"/>
+								</c:if>
 					</c:otherwise>
 				</c:choose>
 				<c:if test="${errorCode != null}">


WebRoot/public/login.jsp
```diff
@@ -459,6 +459,13 @@ if (application.getInitParameter("offlineMode") != null) {
 					<c:otherwise>
 								<form method="post" name="loginForm" action="${formActionUrl}" accept-charset="UTF-8">
 								<input type="hidden" name="loginOp" value="login"/>
+
+                                <c:choose>
```
tarangkhandelwal: small: should this be aligned wit the above input tag?

silentsakky: thanks, fixed

WebRoot/public/login.jsp
```diff
@@ -459,6 +459,13 @@ if (application.getInitParameter("offlineMode") != null) {
 					<c:otherwise>
 								<form method="post" name="loginForm" action="${formActionUrl}" accept-charset="UTF-8">
 								<input type="hidden" name="loginOp" value="login"/>
+
+								<c:choose>
+									<c:when test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">
```
tarangkhandelwal: is it better to use c:if here? as we are only having a single condition ?

silentsakky: changed

WebRoot/public/login.jsp
```diff
@@ -459,6 +459,13 @@ if (application.getInitParameter("offlineMode") != null) {
 					<c:otherwise>
 								<form method="post" name="loginForm" action="${formActionUrl}" accept-charset="UTF-8">
 								<input type="hidden" name="loginOp" value="login"/>
+
+                                <c:choose>
+                                    <c:when test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">
```
komalkaks: Why are we remembering the flag in case of TWO_FACTOR_AUTH_FAILED? 

Suggestion - Break the comment into multiple lines.

silentsakky: yes in case of user has entered wrong otp once, in second attempt if he enters correct otp then we need to maintain rememberme flag till the time authentication gets successful

I will break the comment when squashing commits and merging PR

fatkudu:
Just a general comment.  Are Eastern Shore Library happy to have their name mentioned in a public commit comment?  In general, I think it is best to keep customer names out of git comments.

WebRoot/public/login.jsp
```diff
@@ -460,12 +460,10 @@ if (application.getInitParameter("offlineMode") != null) {
...
 
-								<c:choose>
-									<c:when test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">
-										<!-- if user has selected remember me in login page and we are showing totp screen to user, then we need to maintain value of that flag as after successfull two factor authentication we will have to rewrite ZM_AUTH_TOKEN with correct expires headers -->
-										<input type="hidden" name="zrememberme" value="${param.zrememberme}"/>
-									</c:when>
-								</c:choose>
+								<c:if test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">
+									<!-- if user has selected remember me in login page and we are showing totp screen to user, then we need to maintain value of that flag as after successfull two factor authentication we will have to rewrite ZM_AUTH_TOKEN with correct expires headers -->
+									<input type="hidden" name="zrememberme" value="${param.zrememberme}"/>
```
SatyamKumarOnline: Can we put this code 
<input type="hidden" name="zrememberme" value="${param.zrememberme}"/>

under this condition which is used down in line 473 

<c:when test="${totpAuthRequired || errorCode eq 'account.TWO_FACTOR_AUTH_FAILED'}">

silentsakky: I initially decided to put it there only, but we are already using a hidden input field here so decided to club both input fields instead of putting it into below code which has a proper table structure followed

silentsakky:
@fatkudu thanks will take care of it in future, but I see that the names are appearing in bugzilla ticket also https://bugzilla.zimbra.com/show_bug.cgi?id=107105

fatkudu:
As with most escalations, https://bugzilla.zimbra.com/show_bug.cgi?id=107105 is not publicly visible...

silentsakky:
@fatkudu wasn't aware about it, thanks, I have updated commit messages

PR #42, opened 2017-05-25
ZCS-1315 Fix HTML client broken image issue due to change in universal-ui

silentsakky:
zm-web-client
- All svg files have been moved to it's individual svg folder, and all img files are restored back so html client can use those
- build.xml is modified to do spriting of svgs in a separate folder which we can include in webclient
- all jsp files are modified to include svgs.css file which contains css selectors to use svg using background property
- removed ajax.build.dir variable from build.xml as it is not used anymore
- removed taglib-jar ant target as we are not generating jar for taglib from webclient, it is handled by ivy now
- removed admin.jsp as that is part of admin console

zm-ajax
- modify SkinResources.java to serve svgs.css file if requested by jsp pages
- modified AjxImg.js to add extra prefix in case of svg, so if we are using svg to render an image, the svg tag will have imagename with -svg postfix, this will make sure that we don't have name conflicts between normal and vector images, as vector images will always be reffered using -svg postfix, and normal images will be reffered using only image name
- cleanup code in AjxImg.js to make sure we are not applying same class name to multiple elements, this makes targetting specific elements easier

ZCS-1315 Fix HTML client broken image issue due to change in universal-ui
- dwt.css --> removed transparent border around DwtCalendarButtons which actually makes button flicker on hover, this was not needed
- dwt.css, zm.css --> append -svg to classes which are used with svg images, as normal image name will refer to png/gif images
- ZmListView.js, ZmTagsHelper.js --> changed class names which are used for wrapper element of check/uncheck images, this will resolve conflict of using same class in image div and wrapper div
- skin.css --> Always hide ImgBlank_16 elements as these elements are of no use and creates issue by changing height/width of elements
	--> change path of zimbra.svg file as not it is moved to svg folder


WebRoot/skins/_base/base4/skin.css
```diff
@@ -695,181 +698,181 @@ SELECT                      {   @FieldBorder@   }
...
     @FileTypeMail@
 }
 
-.svg-icon.ImgHtmlDoc {
+.ImgHtmlDoc-svg .svg-icon {
     @FileTypeHtml@
 }
 
-.svg-icon.ImgContactDoc {
+.ImgHtmlDoc-svg .svg-iconx {
```
tarangkhandelwal: typo? should be .svg-icon I think.

.gitignore
```diff
@@ -1,10 +1,3 @@
 build/
 .project
-/WebRoot/js/zimbra/
```
kalpeshhpatel: Do we really need to remove these .gitignore entries?

build.xml
```diff
@@ -1358,11 +1427,8 @@ ext = BeanUtils.cook(ext);
     <target name='server-war'>
         <ant dir='${server.dir}' target='war'/>
     </target>
-    <target name='taglib-jar'>
```
kalpeshhpatel: Why do we need to remove this target `taglib-jar`? I also saw some taglib related changes above, do they really need here?

Thanks

.gitignore
silentsakky: Actually it is removed, as now ant clean will cleanup these files, it was not doing that earlier

So thought that it would be better to remove it

WebRoot/skins/_base/base4/skin.css
silentsakky: indeed it was, thanks for pointing out

fixed now

build.xml
silentsakky: It was part of cleanup work, we don't need to generate taglib-jar as part of webclient code
it is handled by ivy now

silentsakky:
> why is overlay used in svg's? I am not very sure if we still need that for Img's too.
Even I am not sure why it is there, but did't want to break any integration related stuff so I haven't touched it

silentsakky:
> target looks very similar to images target.Is it possible to somehow use a common function and pass param to that ? (images/svgs)
> target looks very similar to images-skins target.Is it possible to somehow use a common function and pass param to that ? (images/svgs)

Both these targets are only defining some paths/variables that can be used in images-merge (which is common for svg and images) target, not sure we need a top level target which does the same thing here

.gitignore
kalpeshhpatel: Okay. No problem. My concern was, `.gitignore` is also helpful to us during development, so that when we create build locally, newly create build and files added during build creation don't shows up untracked files in git log, which is confusing for us. 

Current version of `.gitignore` helps to avoid this by ignoring all files created/added during build time and to make sure that git log only shows actually files which are potential candidate to stage and commit.

build.xml
kalpeshhpatel: Okay. Thanks

.gitignore
silentsakky: Added it back :)

build.xml
```diff
@@ -70,10 +66,8 @@
 	<property name='jsapi-inventory.file' value='${build.docs.dir}/jsapi-inventory.zip'/>
 	<property name='jsapi-changelog.file' value='${build.docs.dir}/jsapi-changelog.zip'/>
 
-    <property name='build.img.dir' value='${build.web.dir}/img'/>
-	<property name="build.css.file" value="images"/>
-	<property name="build.css.ext" value=".css"/>
-    <property name="build.css.filename" value="${build.css.file}${build.css.ext}"/>
+    <property name="build.css.ext" value=".css"/>
+    <!--property name="build.css.filename" value="${build.css.file}${build.css.ext}"/ -->
```
tarangkhandelwal: can we remove this?

silentsakky: we have commented some code in below part also which requires this variable so I have commented this also

we can remove in future

PR #53, opened 2017-06-06
ZCS-1168: Universal UI- New task view

kalpeshhpatel:
Changes:

* Added/updated required CSS for `dwt.css`, `zm.css`, `skin.css`
* Added/updated required strings for `skin.properties`
* Added/updated required icon
* Added/updated templates as per requirement for `Appointment.template` and `Tasks.template`
* ZmTaskEditView,js:
  * Updated template to render for task edit/create
  * Updated icons for priority dropdown
  * Removed inline styles for widgets so that it can be styled by css
* ZmCalItemEditView.js
  * Removed inline styles for widgets so that it can be styled by css
* ZmCalendarApp.js
  * Updated icon for date widget dropdown.

https://jira.corp.synacor.com/browse/ZCS-1168


WebRoot/css/zm.css
```diff
@@ -1257,6 +1268,13 @@ div.AttachmentImgContainer {
     vertical-align: top;
 }
 
+.Conv2MsgHeader .info .ZWidgetTable TD.ZDropDown,
+.MsgHeaderTable .info .ZWidgetTable TD.ZDropDown,
+.InvHeaderDiv .ZWidgetTable TD.ZDropDown {
+	@TransparentBg@;
+	border:none;
```
tarangkhandelwal: some property name related to headerDropdown?

WebRoot/css/zm.css
```diff
@@ -2176,7 +2168,42 @@ div.CompDragArea * {
...
+}
+
+.ZmTaskEditHeader .ZmTaskEditHeaderTable > tbody > tr > td {
+	@TaskEditHeaderTableCell@
+}
+
+.ZmTaskEditHeaderTable .ZmFieldsetContent td,
+.ZmTaskEditHeaderTable .ZmFieldsetContent div,
+.ZmTaskEditHeaderTable .ZmFieldsetContent label {
+	@TaskEditViewFontSize@
```
tarangkhandelwal: can't we just set font size for .ZmTaskEditHeaderTable .ZmFieldsetContent? 
Can we have some container defining property name in case someone want's to add more css rule to same selectors not related to font size?

WebRoot/templates/calendar/Appointment.template
```diff
@@ -239,6 +239,7 @@
 
 <!-- NOTE: Hopefully the EditView template can go away as soon as task -->
 <!-- NOTE: compose is redesigned to be more like appt compose. -->
+<!-- TODO: Edit View template for Appointment and Task are now separated. Remove conditional code for task-edit.-->
```
tarangkhandelwal: can we have a ticket to do this so that we do not forget doing this?

WebRoot/css/zm.css
kalpeshhpatel: Actually this should be the case. But we can't do that, because we have CSS declarations somewhere that targets each element explicitly to set font size. So to override those rules, we have to declare CSS like this.

Regarding container, we already have container for `FieldSetContent`

Let me know if you have any questions.

WebRoot/templates/calendar/Appointment.template
kalpeshhpatel: I guess we already have ticket in pipeline for this. When we start working on appointment create form, we will have to visit this code.

WebRoot/css/zm.css
kalpeshhpatel: Done: https://github.com/Zimbra/zm-web-client/pull/53/files#diff-33f9c57f22cd743824b2918132c5d73bR1274

WebRoot/templates/calendar/Appointment.template
kalpeshhpatel: Linked ticket number in comment: https://github.com/Zimbra/zm-web-client/pull/53/files#diff-60a45c724f7aa5b44add9c6b951de7d2R242

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -626,13 +630,14 @@ TaskbarItem-selected            = @ToolbarColor@
 Select                      = @SelectHeight@ @ActiveCursor@
 SelectHeight                = @WidgetHeight@
 SelectSpacing               =
-SelectPadding               = 5px
+SelectPadding               = @WidgetInternalPadding@
+SelectIcon                  = padding: 0 @SelectPadding@;
 SelectTitle                 = padding:0 @SelectPadding@; text-align:left;
 SelectDropDown              = background:@WidgetBgColor@; @LeftSeparator@; padding:0 @SelectPadding@;
-SelectBorder                = @TextAreaBorder@ @MediumRoundCorners@
+SelectBorder                = @TextAreaBorder@ @MediumRoundCorners@ background:@white@;
```
silentsakky: @AppC@?, not exactly sure

kalpeshhpatel: Though `@appc@` value is white, but I don't feel it should be used. We need `white` color only. `@appc@` may be of different color in some child theme.

WebRoot/css/zm.css
tarangkhandelwal: can't we override the individual declarations that target these? (if possible, otherwise will have to go with this only)

kalpeshhpatel: Here is the one of the rule that specifies font-size of each elements,  https://github.com/Zimbra/zm-web-client/blob/develop/WebRoot/css/common.css#L86. 
(Which is very bad IMO)

So to override that, we have to have specific rule for each elements as well.

Let me know if you have any questions.

tarangkhandelwal: If that's the case then this looks fine for now then. Thanks for the details.

PR #59, opened 2017-06-15
ZCS-1365: Bug 107925 - Persistent XSS - snippet [CWE-79]

tarangkhandelwal:
Issue: Shortcut key “q” that triggers to have mail contents in the tooltip is not XSS safe and set’s the content without encoding.

Changeset:
* ZmMailListController.js: Html encoding the mail fragment before setting it to the ui.


WebRoot/js/zimbraMail/mail/controller/ZmMailListController.js
```diff
@@ -387,6 +387,7 @@ function(actionCode, ev) {
 					} else {
 						frag = item.fragment ? item.fragment : ZmMsg.fragmentIsEmpty;
 						if (frag != "") { lv.setToolTipContent(AjxStringUtil.htmlEncode(frag), true); }
```
tarangkhandelwal: This looks better, Thanks.
Updated PR: https://github.com/Zimbra/zm-web-client/pull/59/commits/1fe5ff5d46b408ee52263b547986021f984dc6ef

PR #70, opened 2017-06-30
ZCS-1659 Use TzMsg.properties file for localization of timezone strings

silentsakky:
- Load TzMsg.properties file also where we are loading AjxMsg.properties
- Removed timezone related strings from AjxMsg.properties file
- Removed admin.jsp as it is related to admin console


silentsakky:
Created separate commit for removal of admin.jsp

PR #71, opened 2017-07-03
ZCS-1224 Universal UI - New / edit contact view upload modal visual update

tarangkhandelwal:
ZCS-1224 Universal UI - New / edit contact view upload modal visual update

Description: Wrapping the default browser input file element to just show a custom label element.

Changeset:
 * ZmUploadDialog.js:
	* Added <label> with the input file element.
	* Renamed _handleFileSize to _handleFileSelection to take care of file name & sizes.
	* If single file is selected then file name & size is shown, in case of multiple files the count & total size is shown.
 * ZmMsg.properties: Removing unrequired properties & adding new property for upload button label.
 * zm.css & skin.properties: Adding styling rules for the upload label.


WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
```diff
@@ -55,6 +55,7 @@ ZmUploadDialog.prototype.constructor = ZmUploadDialog;
 
 ZmUploadDialog.UPLOAD_FIELD_NAME = "uploadFile";
 ZmUploadDialog.UPLOAD_TITLE_FIELD_NAME = "uploadFileTitle";
+ZmUploadDialog.UPLOAD_INPUT_LABEL = "ZmUploadFileLabel";
```
kalpeshhpatel: Minor: Can we update property name to indicate that it's `class` for label. I thought initially that it's text of label. 
Probably: `UPLOAD_LABEL_CLASS `

tarangkhandelwal: Updated : https://github.com/Zimbra/zm-web-client/pull/71/commits/930bfc9a421c260dd1c87516ede226f68617b3a5

kalpeshhpatel: Thanks 👍

WebRoot/css/zm.css
```diff
@@ -829,6 +829,14 @@ div.AttachmentImgContainer {
 	@ActiveCursor@
 }
 
+.ZmUploadFileLabel {
+	@uploadFileLabel@
```
silentsakky: minor: first character of property name should be capital

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
```diff
@@ -407,29 +407,27 @@ ZmUploadDialog.prototype._addFileInputRow = function(oneInputOnly) {
...
-
-    var cellLabel = row.insertCell(-1);
-    cellLabel.innerHTML = ZmMsg.fileLabel;
-
 	var cell = row.insertCell(-1);
 	// bug:53841 allow only one file upload when oneInputOnly is set
 	cell.innerHTML = [
-		"<input id='",inputId,"' type='file' name='",ZmUploadDialog.UPLOAD_FIELD_NAME,"' size=30 ",(this._supportsHTML5 ? (oneInputOnly ? "" : "multiple") : ""),">"
+		"<input id='",inputId,"' type='file' name='",ZmUploadDialog.UPLOAD_FIELD_NAME,"' size=30 ",(this._supportsHTML5 ? (oneInputOnly ? "" : "multiple") : ""),
+		" style='width: 0.1px;height: 0.1px;opacity: 0;overflow: hidden;position: absolute;z-index: -1;'>",
```
silentsakky: minor: if possible can we move this to css or skin.properties?

tarangkhandelwal: intentionally kept here just because we are keeping the visibility 0 for this element and showing the label element (styling via css). 
We don't intend to allow input styling customisable or overridden.

WebRoot/css/zm.css
tarangkhandelwal: Fixed here: https://github.com/Zimbra/zm-web-client/pull/71/commits/3cc17aeeeeee4e80192e44eadb691d0fc3f5b580

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
```diff
@@ -479,28 +477,37 @@ ZmUploadDialog.prototype._addFileInputRow = function(oneInputOnly) {
...
 
-ZmUploadDialog.prototype._handleFileSize =
-function(inputEl, sizeEl){
+ZmUploadDialog.prototype._handleFileSelection =
+function(inputEl, fileDetailEl){
 
     var files = inputEl.files;
-    if(!files) return;
+    if(!files || !files.length){
+        fileDetailEl && (fileDetailEl.innerHTML = "");
```
SatyamKumarOnline: We intent to use equal operator, I believe for check

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1153,6 +1153,8 @@ FormButtonWidth              = width:@FormAddRemoveButtonSize@;
 FormRowSpacing               = 0.75em;
 FormAddRemoveButton          = @FormButtonHeight@
 FormButtonStyle              = @cssShadow@:0px 0px 0px transparent; padding:5px; @ActiveCursor@ @PanelBorder@ @MediumRoundCorners@ background-color:@WidgetBgColor@; @ButtonText@
+UploadFileLabel              = color: @PrimaryBrandingColor@; cursor:pointer;
```
SatyamKumarOnline: spacing after color can be removed

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
```diff
@@ -479,28 +477,37 @@ ZmUploadDialog.prototype._addFileInputRow = function(oneInputOnly) {
...
-	    var aCtxt = ZmAppCtxt.handleWindowOpener();
+        var aCtxt = ZmAppCtxt.handleWindowOpener();
         if(size > aCtxt.get(ZmSetting.DOCUMENT_SIZE_LIMIT))
             className = "RedC";
         totalSize += size;
     }
 
-    if(sizeEl) {
-        sizeEl.innerHTML = "  ("+AjxUtil.formatSize(totalSize, true)+")";
+    if(files.length > 1) {
```
SatyamKumarOnline: files.length can be cached in a single variable as it is used couple of times.

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
tarangkhandelwal: Not checking, we are resetting the value  here.

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
tarangkhandelwal: I don't think there's any gain with that. The files object is in scope and so does the property length on that. It would be unnecessary creating another variable here.

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: e88960076e3c62cb22684a1b151cd0c85cb8bd64

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
SatyamKumarOnline: The scope of the newly created variable will be local to function.
But, we if create a variable and store files.length in that, the browser need not query unnecessary file object to get length property from that.
I feel it's good to have it.

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
SatyamKumarOnline: Ohh, yes.
I got this after posting comment.

WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
tarangkhandelwal: As per my opinion it does not add any value . The files object is in the scope and all the properties on that object too (files in our case). 
Adding other team members for their thoughts.
@silentsakky  @kalpeshhpatel

SatyamKumarOnline: It's not about scope but caching the object property into a variable as files.length is used multiple times. It's a good practice to cache object property into a local variable if they are used multiple times.
Please refer :---

https://gamealchemist.wordpress.com/2016/04/15/writing-efficient-javascript-a-few-tips/

Inline content from this page :--

_Best example of such habit is to cache a repeated computation into a var : it will run faster, require less typing / less reading, and when hovering your mouse over the var in your debugger, you can check the cached value at once when debugging…

Rule is to cache everything you’re going to use again into a local var._

tarangkhandelwal: What I understand from the link is it can lead to performance improvement with caching the required object/array, in cases where we are using nested objects or objects having arrays.

In my case I don't have a nested structure anywhere, I am not convinced that caching a **property**  present on the object into a new variable would serve any purpose.!!

As far as debugging is concerned, hovering over the arrayWhat I understand from the link is it can lead to performance improvement with caching the required object/array, in cases where we are using nested objects or objects having arrays.

In my case I don't have a nested structure anywhere, I am not convinced that caching a **property**  present on the object into a new variable would serve any purpose.!!

As far as debugging is concerned, hovering over the array/object shows up all the properties/elements.
There's no rule defined that says Cache everything!!!

silentsakky: are we talking about performance impact or code readability here?

SatyamKumarOnline: This is  good for both performance impact or code readability. 
It's a good practice to have caching, I feel.

kalpeshhpatel: Well, in my opinion, we don't need separate variable here. We are just checking object property (which would be O(1) time - https://stackoverflow.com/a/3859150/1044026)

However need of separate variable will makes sense when we may call object's method or some other function call.

Thanks

SatyamKumarOnline: Caching array length, in this case is a gain.
https://dzone.com/articles/javascript-shoud-i-have-cache

I would have cached that.

Thanks

tarangkhandelwal:
Squashed and merged

PR #73, opened 2017-07-05
ZCS-1210: Universal UI - Update form elements, tables and buttons appearance

kalpeshhpatel:
Changes:
* Added custom styles for checkbox and radio button globally
* Updated header styles for all preference sections
* Added two variant of button styles viz. `ZInlineButton` and `ZAltButton`
* Refactored some properties in `skin.properties` to remove dependency between apps
* Updated templates for preference sections as per requirement.

https://jira.corp.synacor.com/browse/ZCS-1210


WebRoot/css/zm.css
```diff
@@ -3757,7 +3792,11 @@ HTML>BODY .appt-selected .appt_allday_body
 
 /* Table for a section of prefs (head + body) */
 .ZOptionsSectionTable {
-	margin-bottom:1.5em;
+	@PrefSectionContentTable@
+}
+
+.ZOptionsSectionTable * {
+	@PrefViewFontSize@
```
SatyamKumarOnline: Where is this property used. 
I donot see it at other places

WebRoot/css/zm.css
```diff
@@ -3742,7 +3739,45 @@ HTML>BODY .appt-selected .appt_allday_body
...
+
+.ZOptionsField .ZPropertySheet > tbody > tr > td {
+	@PrefRowInlineTable@
+}
+
+.ZOptionsField .ZPropertySheet > tbody > tr > td:first-child {
+	padding-left: 0;
+}
+
+.ZOptionsField .ZPropertySheet > tbody > tr > td:lst-child {
```
tarangkhandelwal: last-child

WebRoot/css/zm.css
SatyamKumarOnline: It's there, Large diffs are not rendered by default so skin.properties was not loaded

WebRoot/css/zm.css
kalpeshhpatel: Thanks for catching this :)

kalpeshhpatel: FIxed: https://github.com/Zimbra/zm-web-client/pull/73/commits/2a5e3718c338ec78ca1a93073cce8c576a414d76#diff-33f9c57f22cd743824b2918132c5d73bR3774

WebRoot/css/zm.css
```diff
@@ -3757,7 +3792,11 @@ HTML>BODY .appt-selected .appt_allday_body
 
 /* Table for a section of prefs (head + body) */
 .ZOptionsSectionTable {
-	margin-bottom:1.5em;
+	@PrefSectionContentTable@
+}
+
+.ZOptionsSectionTable * {
```
silentsakky: do we need a * selector here, i mean if we assign font size to .ZOptionsSectionTable then wouldn't all children of that element inherit the font size?

WebRoot/css/zm.css
```diff
@@ -6007,10 +6008,8 @@ DIV .SuggestBtn {
 	width:500px;
 }
 
-.ZmFolderPropertiesDialog-container .ZSelected .ZTabBorder {
-	background-color:@lighten(AppC,25)@;
-	border-bottom-color:@lighten(AppC,25)@;
-	/*Color comes from WindowInnerBorder */
+.ZmFolderPropertiesDialog-container .ZTabBar {
+	background: @white@;
```
silentsakky: should this be AppC?

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -564,6 +567,47 @@ ButtonLeftIcon                  = margin: 0; text-align:center;
...
+
+AltButtonBorder                    = padding:0 12px; border:none; @MediumRoundCorners@;
+AltButtonBorder-normal             = background:@AltButtonBg@; box-shadow:0 0 1px 1px rgba(0,0,0,0.2);
+AltButtonBorder-disabled           = background:@AltButtonDisabledBg@; box-shadow:none; opacity:0.6;
+AltButtonBorder-hover              = background:@AltButtonActiveBg@
+AltButtonBorder-active             = @AltButtonBorder-hover@
+AltButtonBorder-focused            = @AltButtonBorder-hover@
+AltButtonBorder-selected           = @AltButtonBorder-hover@
+
+AltButtonText                      = color:@AltBtnTextColor@; text-transform:uppercase; font-weight:300; @FontSize-normal@;
```
silentsakky: do we need height/width as auto here? isn't inline-block will take it's height/width as per content size?

WebRoot/css/zm.css
kalpeshhpatel: Yes we need `*` selector here. Because in `common.css` we do have following rules and needs to override it.

```
P, TH, TD, DIV, SELECT, INPUT[type=text], INPUT[type=password], INPUT[type=file], TEXTAREA, BUTTON {
	@FontSize-normal@
}
```

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: We are overriding `height` and `width` here. Yes inline block can take space according to content but we can also specify fix value if needed. Demo: http://karenmenezes.com/inlineblockelements/

Thanks

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -2111,3 +2167,63 @@ CircularProgressBarTextColor        = @TxtC@;
...
+PrefSectionHeaderActions    = @PrefSectionHeaderWrapper@
+PrefSectionContentTable     = margin:0 0 @PrefViewWrapperPadding@ 0;
+PrefSectionContent          = padding:0; margin:0;
+PrefSectionInternalSpacing  = 4px
+PrefSectionRowCell          = padding:@PrefSectionInternalSpacing@ 0;
+PrefSectionRowLabelWrapper  = width:30%; min-width:200px; @PrefSectionRowCell@;
+PrefSectionRowValueWrapper  = width:70%; @PrefSectionRowCell@;
+PrefSectionRowLabel         = @Text@; width:150px; white-space:nowrap; vertical-align:middle; line-height:20px;
+PrefRowCheckWidget          = padding:2px 0;
+PrefRowSelectWidget         = min-width:280px; width:auto; max-width:100%;
```
silentsakky: do we need width: auto here?

WebRoot/css/zm.css
kalpeshhpatel: Updated: https://github.com/Zimbra/zm-web-client/pull/73/commits/57d3b3c1b6908012902f03528496cf577ac5cc5a#diff-33f9c57f22cd743824b2918132c5d73bR6012

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: Well not actually. Just added for semantic reason. As we have use case as follows:

1. Dropdown should be of atleast 280px wide
2. If it's content is more than 280px wide, it's width should be auto
3. However width should not exceeds 100% of containers width

I can remove `width:auto` here. Let me know.

Thanks

WebRoot/css/zm.css
```diff
@@ -3742,7 +3739,45 @@ HTML>BODY .appt-selected .appt_allday_body
 }
 
 .ZOptionsSectionMain {
-	@roundCorners(0px 0px 8px 8px)@
+	@PrefSectionContent@
+}
+
+.ZOptionsSectionMain>TABLE>TBODY>TR>TD {
```
SatyamKumarOnline: Spacing

WebRoot/js/zimbraMail/prefs/view/ZmPreferencesPage.js
```diff
@@ -674,6 +674,9 @@ function(parentIdOrElem, text, width, listener, id) {
 	button.setText(text);
 	button.addSelectionListener(listener);
 	this._replaceControlElement(parentIdOrElem, button);
+	if(className) {
```
SatyamKumarOnline: spacing

SatyamKumarOnline: I think, this if condition can be removed as Dwt.delClass function doesnot do anything if no className param is passed in addClassName function

WebRoot/css/zm.css
kalpeshhpatel: Fixed: https://github.com/Zimbra/zm-web-client/pull/73/commits/1acb3b20efbd59bbaf0676c5bce106017ffa5f22#diff-33f9c57f22cd743824b2918132c5d73bR3745

WebRoot/js/zimbraMail/prefs/view/ZmPreferencesPage.js
kalpeshhpatel: Fixed: https://github.com/Zimbra/zm-web-client/pull/73/commits/1acb3b20efbd59bbaf0676c5bce106017ffa5f22#diff-89dcba3676fdb8b4880ed597ed2c870dR677

kalpeshhpatel: Will you please elaborate more on this? I didn't get your point.

silentsakky: we should check for existence of property before calling addClassName so the if condition is valid here

SatyamKumarOnline: addClassName internally calls DwtControl.prototype.addClassName --> Dwt.addClass  --> Dwt.delClass.
Now, there in Dwt.delClass, the code
`if (!del && !add) { return; }` makes sure that if the className param which is passed in addClassName function is empty, it will not add any class.
So, above if is just an extra and unnecessary check and can be removed.

kalpeshhpatel: Thanks for your input. But I feel that's an interval working of api `addClassName` which is very well abstracted from us.

Now here, we need to add class to `DwtControl` i.e button in our case. We don't have to remove class. So in my opinion `DwtControlInstance.addClassName("className")` api will be more appropriate and readable than `delClass` api. What do you think?

Also if you see, signature of `Dwt.delClass`, it needs DOM element as input. And here at this point, we have just created control, we have not injected in into DOM.

Let me know if you have any more question on this.

Thanks

SatyamKumarOnline: As stated above, if the inner functions call stack of `DwtControlInstance.addClassName` are taking care of condition when no classname is passed, adding a if check is  redundant here, I think.

`Dwt.delClass` is called internally in the subsequent function call stack when ever `DwtControlInstance.addClassName` is called and I referred them above for reference.

tarangkhandelwal: If its about the initial comment that Dwt.delClass doesn't do anything if no params are passed, my opinion is that it doesn't harm to have a check as className could resolve to null/undefined as well. 

Suggestion: 
There could be another way to write the same code:
className && button.addClassName(className);

kalpeshhpatel: Agreed to your point. However `if` condition also ensures here that is `className` is undefined, let's not call `addClassName` at all. Otherwise if we remove `if(className) {}`, then every time it will stack into `addClassName => Dwt.addClass => Dwt.delClass` calls, even if `className` is undefined and will return from `Dwt.delClass` method. With this `if`,  we are directly avoiding this when it's not needed at all.

Let me know your thought.

Thanks

kalpeshhpatel: @tarangkhandelwal, thanks for your suggestion regarding refactoring here. I did same initially. However looking at line: https://github.com/Zimbra/zm-web-client/pull/73/files/57d3b3c1b6908012902f03528496cf577ac5cc5a#diff-89dcba3676fdb8b4880ed597ed2c870dR669 , I changed it to like this to be consistent here. 

Thanks

SatyamKumarOnline: Sound fine.

PR #74, opened 2017-07-06
ZCS-1225 :- Universal UI - Delegate permisions modal visual update

SatyamKumarOnline:
Description :- Changed style and layout of grant right dialog to accomodate
 the design for new modal.

Changeset:-
a) skin.properties
b) Pages.template
c) ZmAccountsPages.js

### DIFF
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -1667,7 +1667,27 @@ div.CompDragArea * {
 
 .ZmGrantRightsDialog .addrBubbleContainer,
 .ZmPreferencesPage .addrBubbleContainer {
-    width: 290px;
+	@GrantRightsDialogInputContainer@
+}
+
+.ZmGrantRightsDialog .WindowOuterContainer {
+	@GrantRightsDialogOuterContainer@
+}
+
+.ZmGrantRightsDialog #ZmGrantRightsDialog_name_row .Label {
+	@GrantRightsDialogEmailLabel@
+}
+
+.ZmGrantRightsDialog INPUT[type='text'] {
+	@GrantRightsDialogInput@
+}
+
+.ZmGrantRightsDialog .SendType td {
+	padding-top:@SkinWrapperPadding@;
+}
+
+.ZmGrantRightsDialog INPUT[type='checkbox'] {
+	@GrantRightsDialogCheckBox@
 }
 
 .addrBubbleContainerPending {
--- a/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
+++ b/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
@@ -3189,6 +3189,8 @@ ZmGrantRightsDialog = function(parent, className, callback) {
 
 	this._addrInputField.reparentHtmlElement(cellId);
 	this._initAutoComplete();
+	var okBtn = this.getButton(DwtDialog.OK_BUTTON);
+	okBtn.setText(ZmMsg.save);
 };
 
 ZmGrantRightsDialog.prototype = new ZmDialog;
--- a/WebRoot/skins/_base/base4/skin.properties
+++ b/WebRoot/skins/_base/base4/skin.properties
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
 CreateContactGroupDialogCell    = padding:0 0 8px 0;
 CreateContactGroupDialogInput   = @FullWidth@
 
+#######################################################################
+#   Grant Rights Dialog
+#######################################################################
+
+GrantRightsDialogOuterContainer     = background-color:@lighten(SecondaryBrandingColor,94)@;
+GrantRightsDialogInputContainer     = width:290px; border-width:1px; @cssShadow@:inset 0 0 0.5px 0 @IconMsgStatus@; @MediumRoundCorners@;
+GrantRightsDialogInput              = padding:0px 0px 0px 8px;
+GrantRightsDialogEmailLabel         = padding-bottom:10px;
+GrantRightsDialogCheckBox           = margin:0px 8px 0px 0px;
+
 PopupWindowBorder           = @MediumRoundCorners@  @PanelBg@  @PopupShadow@
 
 SmallTransparentBorder      = padding:1px; border:1px solid transparent;
--- a/WebRoot/templates/prefs/Pages.template
+++ b/WebRoot/templates/prefs/Pages.template
@@ -2207,14 +2207,15 @@ NOTE: The tables inside of the filter rule sections MUST contain a tbody element
 <template id='prefs.Pages#GrantRightsDialog'>
 	<table role="presentation">
 		<tr id='${id}_name_row'>
-			<td class="Label" style="padding-right:10px"><$=ZmMsg.emailLabel$></td>
+			<td class="Label"><$=ZmMsg.emailLabel$></td>
+		</tr>
+		<tr>
 			<td id="${id}_name_cell">
 			</td>
 		</tr>
 		<tr>
-			<td class="Label">&nbsp;</td>
 			<td>
-				<table role="presentation">
+				<table class="SendType" role="presentation">
 					<tr>
 						<td><input type="checkbox" id='${id}_sendAs'></td>
 						<td><label class="Text" for='${id}_sendAs'


WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
```diff
@@ -3212,6 +3212,7 @@ function() {
     } else {
      this._sendAs.focus();
     }
+	this.getButton(DwtDialog.OK_BUTTON).setText(ZmMsg.save);
```
silentsakky: minor: indentation

silentsakky: Can't we set this text when creating dialog instance?

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
 CreateContactGroupDialogCell    = padding:0 0 8px 0;
 CreateContactGroupDialogInput   = @FullWidth@
 
+#######################################################################
+#   Grant Rights Dialog
+#######################################################################
+
+GrantRightsDialogOuterContainer     = background-color:@lighten(SecondaryBrandingColor,94)@;
+GrantRightsDialogInputContainer     = width:290px; border-width:1px; @cssShadow@:inset 0 0 0.5px 0 @IconMsgStatus@; @borderRadius@:2px;
```
silentsakky: why we are depending on IconMsgStatus ?

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
...
+GrantRightsDialogInput              = padding:0px; padding-left:8px;
```
silentsakky: i would prefer to use 
padding: 0px 0px 0px 8px;

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
...
+GrantRightsDialogEmailLabel         = padding-bottom:10px;
+GrantRightsDialogCheckBox           = margin-left:0px; margin-right:8px;
```
silentsakky: same as above

WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
tarangkhandelwal: I feel it would be better if we would set the label while dialog is being constructed instead of setting it on popup.

WebRoot/templates/prefs/Pages.template
```diff
@@ -2207,22 +2207,24 @@ NOTE: The tables inside of the filter rule sections MUST contain a tbody element
...
+		</tr>
+		<tr>
 			<td id="${id}_name_cell">
 			</td>
 		</tr>
 		<tr>
-			<td class="Label">&nbsp;</td>
+			<td class="Label"></td>
 			<td>
 				<table role="presentation">
```
kalpeshhpatel: I don't feel need to have two separate classes `SendAs` and `SendObo` to `TD`. I guess you have added these class to target an TD, and same styling is needed for TD in this case. In such cases, you could have targeted TD from TABLE as well, so when new row gets added, we don't have to figure out and add class for TD's of new TR.

Also you can have kept same className for TD's as well.

Thanks

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: For `@borderRadius@:2px;`, please use `@MediumRoundCorners@`

Thanks

WebRoot/templates/prefs/Pages.template
```diff
@@ -2207,22 +2207,24 @@ NOTE: The tables inside of the filter rule sections MUST contain a tbody element
...
-			<td class="Label" style="padding-right:10px"><$=ZmMsg.emailLabel$></td>
+			<td class="Label"><$=ZmMsg.emailLabel$></td>
+		</tr>
+		<tr>
 			<td id="${id}_name_cell">
 			</td>
 		</tr>
 		<tr>
-			<td class="Label">&nbsp;</td>
+			<td class="Label"></td>
```
tarangkhandelwal: don't think we need this <td> as other rows are only having single <td> element. This row can be of same type.

WebRoot/css/zm.css
```diff
@@ -1667,7 +1667,31 @@ div.CompDragArea * {
...
+
+.ZmGrantRightsDialog #ZmGrantRightsDialog_name_row .Label {
+	@GrantRightsDialogEmailLabel@
+}
+
+.ZmGrantRightsDialog INPUT[type='text'] {
+	@GrantRightsDialogInput@
+}
+
+.ZmGrantRightsDialog .Label:empty {
```
tarangkhandelwal: ".ZmGrantRightsDialog .Label:empty " selector can be removed if we are removing the empty .Label element.

WebRoot/css/zm.css
```diff
@@ -1667,7 +1667,31 @@ div.CompDragArea * {
...
+	display:none;
+}
+
+.ZmGrantRightsDialog .SendAs, .ZmGrantRightsDialog .SendObo {
+	padding-top:@SkinWrapperPadding@;
```
tarangkhandelwal: don't think its a right use of skin **wrapper** padding value. 
Also better approach could be to apply line-height on the <tr> and not using padding on the <td>'s, this will keep the content vertically center as well.

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
 CreateContactGroupDialogCell    = padding:0 0 8px 0;
 CreateContactGroupDialogInput   = @FullWidth@
 
+#######################################################################
+#   Grant Rights Dialog
+#######################################################################
+
+GrantRightsDialogOuterContainer     = background-color:@lighten(SecondaryBrandingColor,94)@;
```
tarangkhandelwal: Property : "AttachDialogContainer" also has a rule for background color in modal. It would be great if we can create a common property somewhere for modal background and refer that in:

- AttachDialogContainer 

- GrantRightsDialogOuterContainer

This will allow any change in future for bg color to be made in single location.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: IconMsgStatus has value of rgba(0, 0, 0, 0.5), which is same as that is used here.
So used that rather than creating a new one.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: The background color for different modals are different and are not consistent.
Say Attach Dialog has different background-color than  create new task folder, new mail folder,  shortcut, offline modal etc. 
And, it would not be a good idea to keep a generic background color for all the modal.
Attach Dialog and grant right dialog have very much different context in use.
Therefore, have used different property for GrantRightsDialogOuterContainer.

WebRoot/css/zm.css
SatyamKumarOnline: I donot think using line height will make difference, here.
I prefer using padding as it gives  more control across browsers, as line height can vary on which font you are using, along with what fonts are installed/not installed on the clients browser.

WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
SatyamKumarOnline: About indentation, this function has mix of tabs and spaces both, strange !
Fixed setting text  https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Fixed, https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Fixed, https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/templates/prefs/Pages.template
SatyamKumarOnline: Thanks, done.
https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Yes, here it is 
https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/templates/prefs/Pages.template
SatyamKumarOnline: Thought of removing it initially but I kept this thinking it may come handy in some unknown scenario.
Now, removed.
https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/css/zm.css
SatyamKumarOnline: yes. Removed
https://github.com/Zimbra/zm-web-client/commit/51baf34a4d64bdf1793742ac723637ba7c84bb66

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: ok. So in that case we would have one property to set background for each modal we have ? ( we do have a lot of modal's though).  
It might be good to discuss it with UX ?

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1601,6 +1601,16 @@ CreateContactGroupDialogTable   = @FullWidth@
...
+GrantRightsDialogInputContainer     = width:290px; border-width:1px; @cssShadow@:inset 0 0 0.5px 0 @IconMsgStatus@; @MediumRoundCorners@;
+GrantRightsDialogInput              = padding:0px 0px 0px 8px;
+GrantRightsDialogEmailLabel         = padding-bottom:10px;
+GrantRightsDialogCheckBox           = margin-left:0px 0px 0px 0px 8px;
```
tarangkhandelwal: looks like an invalid assignment for margin-left to me!

WebRoot/css/zm.css
tarangkhandelwal: That's fine if that makes more sense here .

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: Agreed, we can have discussion with UX and should make all modals consistent. In that case we should revisit this and all modals to use common properties.

WebRoot/css/zm.css
SatyamKumarOnline: I briefly looked at Gmail and Face book settings page, where the layout is more or less same as ours  and they are using padding to bring similar to this.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Yes, discussion with UX will be good here.
But, what I feel is that  different modals differ in their taste.
Some are functional (ex. attach dialog), others are informative (Key board shortcut ex), others are for grating rights etc.
May be for UX guys, idea is to have some differentiation in the modal  for the same.
Just a thought.
Worth to bring this up with UX.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Ya typ, fixed here.
https://github.com/Zimbra/zm-web-client/commit/0dfbbe6d1787361c6715ad411cbe140a4b90d546

PR #79, opened 2017-07-07
ZCS-1661 Fix failing build

silentsakky:
- use newly created resolver for zip files
- Don't package AjxTimezoneData.js file in Startup_1.1.js as it is loaded by jsp

### DIFF
--- a/WebRoot/js/zimbraMail/package/DocsPreview.js
+++ b/WebRoot/js/zimbraMail/package/DocsPreview.js
@@ -25,7 +25,6 @@ AjxPackage.require("ajax.util.AjxStringUtil");
 AjxPackage.require("ajax.util.AjxTimedAction");
 AjxPackage.require("ajax.util.AjxText");
 
-AjxPackage.require("ajax.util.AjxTimezoneData");
 AjxPackage.require("ajax.util.AjxTimezone");
 
 AjxPackage.require("ajax.net.AjxRpcRequest");
--- a/WebRoot/js/zimbraMail/package/Startup1_1.js
+++ b/WebRoot/js/zimbraMail/package/Startup1_1.js
@@ -49,7 +49,6 @@ AjxPackage.require("ajax.net.AjxInclude");
 AjxPackage.require("ajax.events.AjxListener");
 AjxPackage.require("ajax.util.AjxText");
 AjxPackage.require("ajax.util.AjxDateUtil");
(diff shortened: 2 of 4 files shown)


WebRoot/js/zimbraMail/package/DocsPreview.js
```diff
@@ -25,7 +25,7 @@ AjxPackage.require("ajax.util.AjxStringUtil");
 AjxPackage.require("ajax.util.AjxTimedAction");
 AjxPackage.require("ajax.util.AjxText");
 
-AjxPackage.require("ajax.util.AjxTimezoneData");
+//AjxPackage.require("ajax.util.AjxTimezoneData");
```
tarangkhandelwal: Do we not need AjxTimezonData anymore here?

WebRoot/js/zimbraMail/package/Startup1_1.js
```diff
@@ -49,7 +49,7 @@ AjxPackage.require("ajax.net.AjxInclude");
 AjxPackage.require("ajax.events.AjxListener");
 AjxPackage.require("ajax.util.AjxText");
 AjxPackage.require("ajax.util.AjxDateUtil");
-AjxPackage.require("ajax.util.AjxTimezoneData");
+//AjxPackage.require("ajax.util.AjxTimezoneData");
```
tarangkhandelwal: Do we not need AjxTimezonData anymore here?

WebRoot/js/zimbraMail/package/DocsPreview.js
silentsakky: it is included from jsp file, we can't package this file with other files as it can be updated at runtime

WebRoot/js/zimbraMail/package/Startup1_1.js
silentsakky: same as above

WebRoot/js/zimbraMail/package/DocsPreview.js
tarangkhandelwal: From which jsp this is included? 
Can we remove the commented part if we won't be needing this?

tarangkhandelwal:
Can we break the commits into two?
Something like:  
1. Fixing the breaking build with updating the resolver to use.
2. Removing dependency on static timezone data output

silentsakky:
Separated out commits

PR #88, opened 2017-07-13
ZCS-1199 :- Universal UI - Compose Action bar and options menu

SatyamKumarOnline:
Desc:- Changes to adjust the UI as per the new mock up for https://jira.corp.synacor.com/browse/ZCS-1199.

Changeset :-
1. ZmOperations.js
2. ZmMailApp.js
3. ZmComposeController.js
4. ZmMsg.properties


WebRoot/js/zimbraMail/core/ZmOperation.js
```diff
@@ -185,7 +185,7 @@ function() {
 		AjxCallback.simpleClosure(function(parent) {
 			ZmOperation.addDeferredMenu(ZmOperation.addSearchMenu, parent, true);
 		}));
-	ZmOperation.registerOp(ZmId.OP_SEND, {textKey:"send", tooltipKey:"sendTooltip", image:"Send", shortcut:ZmKeyMap.SEND});
+	ZmOperation.registerOp(ZmId.OP_SEND, {textKey:"send", tooltipKey:"sendTooltip", image:"Send", shortcut:ZmKeyMap.SEND, showImageInToolbar: true, showTextInToolbar: true});
```
silentsakky: in case of we have preference MAIL_SEND_LATER_ENABLED as true

then we are using ZmId.OP_SEND_LATER operation, does this changes needs to be done there as well?

silentsakky: this same operation is used in ZmConvReplyView.prototype._initializeToolbar also, do we need these changes in that view also?

WebRoot/js/zimbraMail/mail/controller/ZmComposeController.js
```diff
@@ -1228,7 +1228,7 @@ function() {
 		buttons.push(ZmOperation.SEND);
 	}
 
-	buttons.push(ZmOperation.CANCEL, ZmOperation.SEP, ZmOperation.SAVE_DRAFT);
+	buttons.push(ZmOperation.SAVE_DRAFT, ZmOperation.SEP, ZmOperation.CANCEL);
```
silentsakky: what does ZmOperation.SEP do here? can we remove if not required
not sure are we maintaining this in other places

@tarangkhandelwal @kalpeshhpatel let me know your thoughts about this

tarangkhandelwal: Its good if we can get rid of ZmOperation.SEP from this toolbar. Though, I see on UI it has also been suppressed so not much effect from visibility perspective but its always good to keep DOM clean.

WebRoot/js/zimbraMail/core/ZmOperation.js
SatyamKumarOnline: I think NO. When the preference MAIL_SEND_LATER_ENABLED as true, we show 'Send Later' as a menu option and not in the toolbar. The above changes deal with showing images/ text in toolbar.

SatyamKumarOnline: Yes, please refer :--
https://app.zeplin.io/project/587bfff6082c2ff97b67ee6a/screen/5890d1a7cb4fd20443b28511

WebRoot/js/zimbraMail/mail/controller/ZmComposeController.js
SatyamKumarOnline: ZmOperation.SEP is creating a bit of vertical spacing between the two labels, here. It can be removed as it is suppressing the spacing between two labels to small extent.
But, there are many other places in the same view , where ZmOperation.SEP is being used and I didn't removed them to be consistent.

WebRoot/js/zimbraMail/core/ZmOperation.js
silentsakky: Thanks

SatyamKumarOnline: Modified the 'overrides' object, here.
https://github.com/Zimbra/zm-web-client/commit/b2c6653e4ee460ed03a11702ea11f17725bac5ca

PR #89, opened 2017-07-13
ZCS-1727: External account shows 500 error when password restrictions not met

kalpeshhpatel:
Changes:

* extuserprov.jsp:
  * Added server-side validation check for password error

https://jira.corp.synacor.com/browse/ZCS-1727

### DIFF
--- a/WebRoot/messages/ZhMsg.properties
+++ b/WebRoot/messages/ZhMsg.properties
@@ -1854,6 +1854,7 @@ account.IDENTITY_EXISTS = Identity already exists.
 account.TOO_MANY_IDENTITIES = An identity could not be created because you have exceeded the limit.
 account.DATA_SOURCE_EXISTS = Data source already exists.
 account.TOO_MANY_DATA_SOURCES = A data source could not be created because you have exceeded the limit.
+account.NO_SHARE_EXISTS = No share exists.
 ####### mail.*
 mail.MAINTENANCE = Mailbox is currently in maintenance mode.
 mail.NO_SUCH_MBOX = Mailbox does not exist.
--- a/WebRoot/public/extuserprov.jsp
+++ b/WebRoot/public/extuserprov.jsp
@@ -28,6 +28,7 @@
 
 <c:set var="client" value="${param.client}"/>
 <c:set var="version" value="${initParam.zimbraCacheBusterVersion}"/>
+<c:set var="validationErrorCode" value="${not empty errorCode ? errorCode : (not empty header['errorCode'] ) ? header['errorCode'] : null}" />
 
 <head><title><fmt:message key="externalUserRegistration"/></title>
 <link rel="stylesheet" type="text/css" href="<c:url value='/css/common,login,zhtml.css'>
@@ -63,11 +64,15 @@
 
 			<form action="/service/extuserprov/" method="post" onsubmit="return checkPasswords();">
 
-			<div id="ZLoginErrorPanel" style="display:none;">
+			<div id="ZLoginErrorPanel" style="${not empty validationErrorCode ? 'display:block': 'display:none'}">
 				<table>
 					<tr>
 						<td><app:img id="ZLoginErrorIcon" altkey='ALT_ERROR' src="dwt/ImgCritical_32.png" /></td>
-						<td id="errorMessage"></td>
+						<td id="errorMessage">
+							<c:if test="${not empty validationErrorCode}"> 
+								<fmt:message bundle="${zhmsg}" key='${validationErrorCode}'/>
+							</c:if>
+						</td>
 					</tr>
 				</table>
 			</div>


WebRoot/public/extuserprov.jsp
```diff
@@ -147,5 +151,10 @@
 		return !isError;
 	}
 </script>
+<% if (request.getAttribute("errorCode") != null) { %>
+	<script>
+		document.getElementById("ZLoginErrorPanel").style.display = "block";
+	</script>
+<% } %>
```
tarangkhandelwal: should we also have else to this? 
if errorCode is null then we should set the element to display none.

kalpeshhpatel: No need of `else` block here. It's already managed here: https://github.com/Zimbra/zm-web-client/blob/854a2da5caef46f5b3f643cdff58ed0163105857/WebRoot/public/extuserprov.jsp#L66

WebRoot/public/extuserprov.jsp
```diff
@@ -67,7 +67,11 @@
 				<table>
 					<tr>
 						<td><app:img id="ZLoginErrorIcon" altkey='ALT_ERROR' src="dwt/ImgCritical_32.png" /></td>
-						<td id="errorMessage"></td>
+						<td id="errorMessage">
+							<% if (request.getAttribute("errorCode") != null) { %>
+								<fmt:message bundle="${zhmsg}" key='${errorCode}'/>
```
silentsakky: any specific reason to use zhmsg bundle? i thought for error messages we are using ZMsg bundle

WebRoot/public/extuserprov.jsp
```diff
@@ -147,5 +151,10 @@
 		return !isError;
 	}
 </script>
+<% if (request.getAttribute("errorCode") != null) { %>
+	<script>
+		document.getElementById("ZLoginErrorPanel").style.display = "block";
```
silentsakky: can we move this to line 66, we are already defining display: none there can we conditionally set it?

WebRoot/public/extuserprov.jsp
kalpeshhpatel: I just referred use of `zhMsg` bundle used in other places of same file: https://github.com/Zimbra/zm-web-client/blob/854a2da5caef46f5b3f643cdff58ed0163105857/WebRoot/public/extuserprov.jsp#L139

So to be consistent, I follow same. 

Let me know if you have more insights on this.

Thanks

silentsakky: no issues in using this bundle just wanted to make sure we are covering all error codes :)

as per PR: https://github.com/Zimbra/zm-mailbox/pull/184/files, i can see server can send two error codes 
1) ACCOUNT_EXISTS, this is already present in ZhMsg bundle
2) NO_SHARE_EXISTS, i think this is new error code added, so we probably don't have it in any bundles

@rohan-ambasta let us know, which other error codes we can get in this jsp file besides above two.

WebRoot/public/extuserprov.jsp
plobbes: Moving this to L66 seems to be a simpler fix and avoids this client side javascript hack altogether, no?

WebRoot/public/extuserprov.jsp
rohan-ambasta: NO_SHARE_EXISTS was added. But you can also get AccountServiceException.INVALID_PASSWORD
In case of any unknown exception you may get ServiceException.FAILURE or TEMPORARILY_UNAVAILABLE

WebRoot/public/extuserprov.jsp
kalpeshhpatel: Sure, that will be better. Let me try that. 
Thanks for highlighting.

kalpeshhpatel: @silentsakky  @plobbes Please find update: https://github.com/Zimbra/zm-web-
client/pull/89/files#diff-567b9b4096d03c05b57d704fed7794f6R66

Thanks

WebRoot/public/extuserprov.jsp
kalpeshhpatel: I have handled message for new errorCode.
https://github.com/Zimbra/zm-web-client/pull/89/files#diff-f88cbfd6558faa3922ca722129e31486R1857

@silentsakky @rohan-ambasta 

Thanks

WebRoot/messages/ZhMsg.properties
```diff
@@ -1854,6 +1854,7 @@ account.IDENTITY_EXISTS = Identity already exists.
 account.TOO_MANY_IDENTITIES = An identity could not be created because you have exceeded the limit.
 account.DATA_SOURCE_EXISTS = Data source already exists.
 account.TOO_MANY_DATA_SOURCES = A data source could not be created because you have exceeded the limit.
+account.NO_SHARE_EXISTS = No share exists.
```
kalpeshhpatel: Feel free to suggest proper error message for this.

We may also need localised strings for this property.

silentsakky: Looks good, for translations we can create a new ticket for time being

kalpeshhpatel: Sure. Created ticket to track progress for now.
https://jira.corp.synacor.com/browse/ZCS-1953

Thanks

WebRoot/public/extuserprov.jsp
plobbes: That link seems invalid.

WebRoot/public/extuserprov.jsp
```diff
@@ -28,6 +28,12 @@
 
 <c:set var="client" value="${param.client}"/>
 <c:set var="version" value="${initParam.zimbraCacheBusterVersion}"/>
+<c:if test="${not empty errorCode}">
+	<c:set var="validationErrorCode" value="${errorCode}" />
+</c:if>
+<c:if test="${not empty header['errorCode']}">
+	<c:set var="validationErrorCode" value="${header['errorCode']}" />
+</c:if>
```
PriteshGajjar008: We can shorthand this as below:
 <c:set var="validationErrorCode" value="${not empty errorCode?errorCode:(not empty header['errorCode'])?header['errorCode']:null />

kalpeshhpatel: Done changes:  https://github.com/Zimbra/zm-web-client/pull/89/commits/48ede47dca7ef3009f357eb000a5dfded0b65018#diff-567b9b4096d03c05b57d704fed7794f6R37

PR #92, opened 2017-07-20
ZCS-1883 Firefox shows extra attribute style="" on mail editor(body area) form

silentsakky:
Issue:
- For non IE browsers we were adding empty style tag for outer most div in message body, which is causing issues in selenium test cases

Resolution:
- apply styles only for IE browser for other browsers don't add any style tag

### DIFF
--- a/WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
+++ b/WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
@@ -1053,8 +1053,6 @@ function(params) {
 	DBG.println(AjxDebug.DBG1, "Use IFRAME: " + this._usingIframe);
 	
 	if (this._usingIframe) {
-		// bug fix #9475 - IE isnt resolving MsgBody class in iframe so set styles explicitly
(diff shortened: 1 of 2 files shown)


WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
```diff
@@ -1063,14 +1061,18 @@ function(params) {
...
-			styles:					inner_styles,
 			noscroll:				!this._scrollWithIframe,
 			posStyle:				DwtControl.STATIC_STYLE,
 			processHtmlCallback:	callback,
 			useKbMgmt:				true,
 			title:                  this._getIframeTitle()
 		};
 
+		// bug fix #9475 - IE isnt resolving MsgBody class in iframe so set styles explicitly
+		if(AjxEnv.isIE) {
```
SatyamKumarOnline: Minor :- space after if

silentsakky: I prefer having a space after if condition
let me know if that drastically affects code readability

PR #101, opened 2017-08-01
ZCS-1211: Universal UI - Reorganized signature interface

kalpeshhpatel:
Updated look and feel of Preferences => Signatures interface as per new mockup.

https://jira.corp.synacor.com/browse/ZCS-1211

### DIFF
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -3783,6 +3783,10 @@ HTML>BODY .appt-selected .appt_allday_body
 	@PrefSectionContent@
 }
 
+.ZOptionsSectionMain .ZmOptionsTable {
+	@PrefSectionOptionTable@
+}
+
 .ZOptionsSectionMain > TABLE > TBODY > TR > TD {
 	@PrefSectionRowCell@
 }
@@ -4007,16 +4011,6 @@ HTML>BODY .appt-selected .appt_allday_body
 .ZOptionsAcctName {
 	margin-right:10px;
 	white-space:nowrap;
-}
-
-.ZmSignatureListView {
-/*	width:180px !important;*/
-	height:100%;
-	background-color:white;
-}
-
-.ZmSignatureListView TD {
-	padding-left:5px;
 }
 
 .ZmFilterRulesView {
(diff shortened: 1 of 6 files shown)


WebRoot/css/zm.css
```diff
@@ -7279,4 +7269,78 @@ input[type="radio"] + label {
...
+
+.ZmSignatureNameWrapper {
+	@PrefSignatureNameWrapper@
+}
+
+.ZmSignatureNameWrapper > label {
+	@PrefSignatureNameLabel@
+}
+
+.ZmSignatureNameWrapper .DwtInputField input[type='text'] {
```
tarangkhandelwal: will we be removing this once zcs-2394 is done?

WebRoot/css/zm.css
```diff
@@ -7279,4 +7269,78 @@ input[type="radio"] + label {
...
+
+.ZmSignatureListWrapper {
+	@PrefSignatureListWrapper@
+}
+
+.ZmSignatureListView {
+	@PrefSignatureListView@
+}
+
+.ZmSignatureListView [class*='Row'] TD {
```
tarangkhandelwal: why not using just .Row ?

WebRoot/templates/prefs/Pages.template
```diff
@@ -758,47 +757,54 @@ and the container element is replaced with that control.
 </template>
 
 <template id="prefs.Pages#SignatureSplitView">
-	<table role="presentation" width='100%'>
+	<table role="presentation" width=100%>
```
tarangkhandelwal: minor, intentionally removed single quotes around 100% ?

WebRoot/templates/prefs/Pages.template
```diff
@@ -758,47 +757,54 @@ and the container element is replaced with that control.
...
+				<div class="ZmSignatureEditorWrapper">
+					<div class="ZmSignatureEditor">
+						<div id='${id}_SIG_EDITOR' tabindex=0></div>
+					</div>
+					<div class="ZmSignatureVCardWrapper">
+						<table role="presentation">
+							<tr>
+								<td class="Label"><$=ZmMsg.signatureVcardLabel$></td>
+								<td class='ZOptionsField'><input id='${id}_SIG_VCARD' tabindex=0 size=30></td>
+								<td><div id='${id}_SIG_VCARD_BROWSE' style='padding-right:5px; padding-left:5px;' tabindex=0></div></td>
```
tarangkhandelwal: inilne styling?

WebRoot/templates/prefs/Pages.template
```diff
@@ -758,47 +757,54 @@ and the container element is replaced with that control.
...
-			<td align='right'>
-				<table role="presentation">
-					<tr>
-						<td class="Label"><$=ZmMsg.signatureVcardLabel$></td>
-						<td class='ZOptionsField'><input id='${id}_SIG_VCARD' tabindex=0 size=30></td>
-						<td><div id='${id}_SIG_VCARD_BROWSE' style='padding-right:5px; padding-left:5px;' tabindex=0></div></td>
-						<td><div id='${id}_SIG_VCARD_CLEAR' tabindex=0></div></td>
-					</tr>
-				</table>
+			<td>
```
tarangkhandelwal: suggestion, should we be using **colspan** in the above td?

WebRoot/templates/prefs/Pages.template
kalpeshhpatel: This is part of vCard interface and has to be taken care in separate ticket only. It's existing markup.

WebRoot/templates/prefs/Pages.template
kalpeshhpatel: No no, not intentionally. Must be mistake. I'll revert back.

WebRoot/templates/prefs/Pages.template
kalpeshhpatel: That won''t work here. We don't want column to span two here. It has to be one. And same as above cell.

tarangkhandelwal: sounds good then..

WebRoot/css/zm.css
kalpeshhpatel: There are many classes not just Row, like Row-focussed, Row-disabled etc. For all such cases we need same styling. So instead of targeting each one separately, this works out well.

WebRoot/css/zm.css
kalpeshhpatel: Maybe not. It not only focusses on width, but also has border from all edges as oppose to normal inputs where we have border from bottom only.

WebRoot/templates/prefs/Pages.template
kalpeshhpatel: Fixed here: https://github.com/Zimbra/zm-web-client/pull/101/commits/0c6226af5526ddfcadc6b8a6805428db18ef767c#diff-953f1692b15ba4b991582d9b98cc3fb4R760

Thanks

WebRoot/css/zm.css
```diff
@@ -7279,4 +7273,78 @@ input[type="radio"] + label {
...
+	@PrefSignatureListRow@
+}
+
+.ZmSignatureListView [class*='Row-selected'] {
+	@PrefSignatureListRowActive@
```
SatyamKumarOnline: I was thinking of having a name PrefSignatureListRowSelected, but this is also ok.

SatyamKumarOnline:
Looks good to me .

WebRoot/css/zm.css
tarangkhandelwal: I think Row class is always applied. I agree that other classes do exist but Row would also be present I feel.

kalpeshhpatel: Yes, Row class would always applies. But based on state, sometimes along with `.Row` other class are also get applied. For instance `Row-Selected` or `Row-Focussed` and in any of these cases, I want this rules to get applied. (I am not sure though if any-of those class will break UI or not. Just that I want to be on safer side :) )

Other way to handle this would be to have my selectors as follows:

```
.ZmSignatureListView .Row TD,
.ZmSignatureListView .Row-focussed TD,
.ZmSignatureListView .Row-selected TD,
.ZmSignatureListView .Row-disabled TD {
   //my rules
}
```

Here I am just avoiding this repetition with this selector. 

Thanks

tarangkhandelwal: Got it. I think you know what you are doing here.
Thanks for the explanation.

PR #104, opened 2017-08-02
Various localization fixes

zx-fulvio:
Various fixes across several localization files.
Mostly unescaped newlines, inconsistent key/value separators, key/value definitions mistakenly pasted inside other values, garbled key definitions.


WebRoot/messages/ZmSMS_fr_CA.properties
```diff
@@ -777,4 +777,4 @@ carrier_USVirginIslands_CentennialWireless.pattern = {0}@cwemail.com
 
 carrier_USVirginIslands_TracFone.label = TracFone
 
-carrier_USVirginIslands_TracFone.pattern = {0}@mmst5.tracfone.com
\ No newline at end of file
+carrier_USVirginIslands_TracFone.pattern = {0}@mmst5.tracfone.com
```
silentsakky: does this make any difference?

silentsakky:
Can you squash commits into single one?

WebRoot/messages/ZmSMS_fr_CA.properties
zx-fulvio: It's inconsistent with the other files, and it makes checking for inconsistencies easier.
Functionally I am not sure it matters.

zx-fulvio:
@silentsakky: I can do that. I pushed several commits just to allow to get a clearer picture of what was being corrected.

WebRoot/messages/ZmSMS_fr_CA.properties
silentsakky: ok thanks

zx-fulvio:
@silentsakky: btw there's a sizable problem with quoting all over the localization files.
Basically in the resources bundle there are keys which have plain literal strings as values, and other keys whose values are actually formats with placeholders following the Java MessageFormat scheme and used in the web interface.
The placeholders (e.g. **{0}** ) are then replaced in the code with variable values, such as e-mail addresses.
Unfortunately the MessageFormat format uses the apostrophe character as a quoting character, so if you want to insert a literal apostrophe, `'`, you have to write two adjacent ones, e.g. `''`. If you just write  one apostrophe it just treats everything up to the next apostrophe, _or to the end of the string_, as a literal sequence. Which means that any placeholder following that lone apostrophe is not replaced with anything at all and displayed as it is. See the yellow text in picture below.
selezione_006
`actionFolderCreated = \u05d4\u05ea\u05d9\u05e7\u05d9\u05d9\u05d4 '{0}' \u05e0\u05d5\u05e6\u05e8\u05d4`
In this specific case the apostrophes were used around `{0}` instead of double quotes therefore preventing the sequence from being expanded.
Far more often (and I'm talking about dozens if not hundreds of strings) it seems as though the strings were localized without any awareness about how to write a literal apostrophe in these particular formats (or that they weren't plain strings at all), resulting in missing apostrophes (removed by the formatter as quotation marks) and/or unexpanded placeholders (because lone apostrophe characters might preserve everything up to the end of the string). This is compounded by the fact that there is no indication whatsoever on whether a localized string contains a format or a plain string. I can infer that a lot are formats through the presence of placeholders and fix those, but I can't know if others which have no placeholders inside will be concatenated to a different one which has them and then passed through a formatter in the code.

This is, of course, off-topic here, but since there are **a lot** of strings across a few repositories to correct I was wondering whether I should prepare pull requests containing one giant commit or several smaller ones. I am not sure which approach would simplify a subsequent review for you.

zx-fulvio:
Commits squashed into one.

silentsakky:
@zx-fulvio I think one single commit would be appropriate if the changes are logically similar

Victor-D:
Hi Fulvio,

Since you're working on translation issue you may want to check this bug :
https://bugzilla.zimbra.com/show_bug.cgi?id=106468

iCalendar notifications subject (Accept, Refused, Cancelled, etc.) are never translated and are always in english when caldav protocol is used. 

Regards
Victor

zx-fulvio:
@silentsakky Thanks. I'll see whether to split into a few different ones for the different cases, even though the issues are pretty similar.

zx-fulvio:
@Victor-D Sorry Victor, I don't think I can help you with that.
As you might have seen I've looked into problems with the actual translations, not with the code providing l10n/i18n. I am a sysadmin by trade, not a Java developer. I might be able to spot the problem, but I lack the time and usually the expertise to provide working code to solve it.
As for your issue, I took the time to look at the mailbox code to try and spot where the problem might lie.
I think, with all the caveats above, but I can't be utterly sure about it, that the code preparing the e-mail for the Caldav notification is the [handleEventRequest function here](https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/dav/resource/ScheduleOutbox.java) . If you look for instance at [this code here](https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/mailbox/calendar/CalendarMailSender.java) which I think is used to prepare e-mail for Calendar events triggered by webmail users, you'll see that the code for the Dav portion is missing a lot of calls used to localize the subject and other portions which are present in the other file. So it does seem to be a problem with the internationalization of the code handling the notification e-mail for caldav events, rather than with the translation of the relevant strings. So messing with the translation files won't help you at all.
Again, I am not a Java developer so I can't provide a fix for that (if that _is_ where the problem is) and I wouldn't have the time to try anyway.
I hope though that the info might be helpful regardless. 👋

zx-fulvio:
Forced push with an additional small fix added.

silentsakky:
Merged to develop

PR #107, opened 2017-08-03
ZCS-1203: Universal UI: Long subject line stuck in the message preview window in conversation mode

kalpeshhpatel:
Fix:
In-case of conversation view, subject’s while-space has been set to `no-wrap` as per current behaviour.


WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1270,7 +1270,7 @@ ConvView                    = @PanelBorder@  @Conv2Bg@; box-sizing: border-box;
 Conv2Bg                     = @AppBg@
 Conv2View                   = height: 100%;  margin: @MailViewGutterSpace@;
 Conv2Header                 = background-color: transparent; overflow: hidden; @ActiveCursor@; width: 100%; max-width100%; padding:@MailViewGutterSpace@ 0; margin-bottom: @MailViewInnerSpacing@;
-Conv2HeaderSubject          = @FontSize-biggest@ ;padding: 0 5px; width:95%;@Text@
+Conv2HeaderSubject          = @FontSize-biggest@ ;padding: 0 5px; width:95%; @Text@; @NoWrap@;
```
silentsakky: do we need an ellipsis?

kalpeshhpatel: I think we can have it. It will need change in other properties as well. Let me try it out and see. 
(This is however what current behaviour is in 8.7)

Thanks

silentsakky: ok if it turns out to be too complex then we can leave it as that is current behaviour

kalpeshhpatel: Here is the updated code: https://github.com/Zimbra/zm-web-client/pull/107/commits/b819c03fe97bb773a72c7f8e99330cd4390df549

silentsakky: Cool thanks

PR #110, opened 2017-08-07
ZCS-2822: Changes to create distribution packages for timezones

shri314:


build.xml
```diff
@@ -662,32 +662,6 @@
         </if>
     </target>
 
-    <target name="timezones">
```
silentsakky: although this will work perfectly in case of creating build
but for development environment we are using prod-deploy target which will be broken with this because after deploying the war we will not be copying the timezone related files

shri314: I think we will need to discuss viable dev solutions for the problem you describe.

PR #112, opened 2017-08-09
ZCS-2450 Style the drag proxy element for Contact list

tarangkhandelwal:
**Fix Description**: 
The new styling changes which includes CSS selectors were targeting elements inside the ListView (ZmContactSimpleView). In case of proxy element, there’s no context of the ListView or its classes & thus partially some rules were not getting applied on the proxy & was breaking the UI. Fix the issue by removing the scope of the list view class & depending only on the Contact row specific class “SimpleContact”.

**Changeset**:
 * zm.css: Updated Row specific style selectors to not depend on the listview  class ZmContactSimpleView  but on the specific class i.e. SimpleContact only
 * ZmContactSplitView.js:
   *  Removing dragProxy specific logic for changing width/padding. Matching the UI element to be same as the list item element.
   *  Changing the order of row class for the Contact list item so that a consistent class name is applied across the apps for dragProxy. (Row-dragProxy)

### DIFF
--- a/WebRoot/js/zimbraMail/abook/view/ZmContactSplitView.js
+++ b/WebRoot/js/zimbraMail/abook/view/ZmContactSplitView.js
@@ -1036,7 +1036,7 @@ ZmContactSimpleView = function(params) {
 	params.className = "ZmContactSimpleView";
 	ZmContactsBaseView.call(this, params);
 
-	this._normalClass = DwtListView.ROW_CLASS + " SimpleContact";
+	this._normalClass = "SimpleContact " + DwtListView.ROW_CLASS;
 	this._selectedClass = [DwtListView.ROW_CLASS, DwtCssStyle.SELECTED].join("-");
 };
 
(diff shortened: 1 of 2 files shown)


WebRoot/js/zimbraMail/abook/view/ZmContactSplitView.js
```diff
@@ -1036,7 +1036,7 @@ ZmContactSimpleView = function(params) {
 	params.className = "ZmContactSimpleView";
 	ZmContactsBaseView.call(this, params);
 
-	this._normalClass = DwtListView.ROW_CLASS + " SimpleContact";
+	this._normalClass = "SimpleContact " + DwtListView.ROW_CLASS;
```
silentsakky: sequence of class should not matter, is this done by mistake?

tarangkhandelwal: It does matter. DragProxy logic appends "-dragProxy" tot he end of the _normalClass variable. Updated the order to have it consistent class name across the apps. (Row-dragProxy/RowDouble-dragProxy).

silentsakky: cool I was thinking in terms of css only

PR #114, opened 2017-08-10
ZCS-2282 Build script changes for deploying timezone files in dev setup

silentsakky:
- fix of https://github.com/Zimbra/zm-web-client/pull/110#discussion_r131870886 introduced a change which broke dev setup
- for dev setup, we will copy timezone files after deploying webapp and extracting it

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -117,7 +117,7 @@
 			<equals arg1="${zimbra.buildinfo.type}" arg2="NETWORK" />
 		</condition>
         <property name='common.web.includes' value='${custom.web.includes},js/**,css/**'/>
-        <property name='common.web.excludes' value='${custom.web.excludes},js/*/package/**,**/AjxTimezoneData.js'/>
+        <property name='common.web.excludes' value='${custom.web.excludes},js/*/package/**'/>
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -1148,9 +1148,26 @@ ext = BeanUtils.cook(ext);
         <unzip dest='${webapp.dir}' src="${webapp.file}"/>
 
         <antcall target='webxml-deploy-replace'/>
+        <antcall target='timezones-deploy'/>
```
tarangkhandelwal: Can we confirm if this change doesn't affect the change done by @shri314  (https://github.com/Zimbra/zm-web-client/pull/110)  ? 
The jam-files target was changed there. With this change in 'deploy' , internally 'jam-files' too get  called. 
Can we confirm if the SE team is not using this target at all and just depending on the war generated ?

silentsakky: build script is using prod-war target
and we are modifying deploy target, i guess deploy is only used for deploying war in dev setup

tarangkhandelwal: ok just wanted to make sure of it. Thanks

PR #115, opened 2017-08-10
Fixing literal apostrophe characters not escaped in format strings

zx-fulvio:
As mentioned in [this comment to a different pull request](https://github.com/Zimbra/zm-web-client/pull/104#issuecomment-320630430) which contains a concrete example, there is a widespread problem with the localization of MessageFormat-like formats ( AjxMessageFormat ) in the resource files.

Basically in the resource files values might contain either a plain string or a format. The latter might or might not contain a placeholder such as {0} which will be replaced in the output with the value of a variable through the use of a formatter in the code.
In the MessageFormat definition, a pair of apostrophe characters (ASCII character 39 / Unicode codepoint U+0027) is used to prevent interpolation of anything between the two. A lone apostrophe will effectively prevent interpolation up to the end of the string. Anyone needing to insert a literal apostrophe in the string would need to write a sequence of two of them (i.e. `''`). And here lies the problem. In a lot (**a lot**) of strings literal apostrophes are written as-is (i.e. `'`), not as `''`.
This means for instance that a value such as `L'alias {0} \u00e8 stato cancellato.` (The alias {0} has been deleted.) in the resource file will not display in the web interface as
`L'alias alias@example.com è stato cancellato.`
as intended but rather as 
`Lalias {0} è stato cancellato.`.
So `'` disappears and `{0}` is preserved as a literal substring.

Unfortunately format strings are not flagged as such in the resource files, so one cannot know beforehand whether a value contains a format or a plain strings unless one reads the related code. And in the code sometimes the strings is used either as a plain string or as a format according to the actual contents of the value.
Therefore I chose values containing a construct such as `{0}` or ChoiceFormats as targets for replacing `'` with `''`, even though I located other values used as formats with no such constructs. In other words it's a partial fix.

Provided that I have not located a single instance where `'` was used as an actual quoting character in the formats rather than its literal self, my suggestion to fix the problem now and in the future would be to _avoid entirely_ that character in the values. There are plenty of non-ASCII Unicode characters which are a better fit (such as [RIGHT SINGLE QUOTATION MARK](http://unicode.org/cldr/utility/character.jsp?a=2019), sometimes present in the values, even intermixed with `'` in the same value) and would prevent any quoting problem from occurring whenever a translator forgets (or just doesn't know) the a specific value does not contain a plain string. But it would up to the translators to know which one to use in this or that language, therefore I won't be making arbitrary substitutions here. It would also simplify dramatically checking for issues since any appearance of  `'` would be suspect or outright incorrect and easy to spot programmatically.

I prepared pull requests for the zm-admin-console and zm-mailbox repositories addressing the same issue.


WebRoot/messages/ZMsg_iw.properties
```diff
@@ -39,18 +39,18 @@ account.PASSWORD_LOCKED = \u05d4\u05e1\u05d9\u05e1\u05de\u05d4 \u05e0\u05e2\u05d
 account.PASSWORD_RECENTLY_USED = \u05d4\u05e1\u05d9\u05e1\u05de\u05d4 \u05e9\u05e9\u05dc\u05d7\u05ea \u05d4\u05d9\u05d9\u05ea\u05d4 \u05d1\u05e9\u05d9\u05de\u05d5\u05e9 \u05dc\u05d0\u05d7\u05e8\u05d5\u05e0\u05d4. \u05d1\u05d7\u05e8 \u05e1\u05d9\u05e1\u05de\u05d4 \u05d0\u05d7\u05e8\u05ea.
 
 # {0} = filename
-formatter.INVALID_FORMAT = \u05ea\u05d1\u05e0\u05d9\u05ea \u05e7\u05d5\u05d1\u05e5 \u05dc\u05d0-\u05d7\u05d5\u05e7\u05d9\u05ea \u05e2\u05d1\u05d5\u05e8 \u05d4\u05e7\u05d5\u05d1\u05e5 '{0}'.
+formatter.INVALID_FORMAT = \u05ea\u05d1\u05e0\u05d9\u05ea \u05e7\u05d5\u05d1\u05e5 \u05dc\u05d0-\u05d7\u05d5\u05e7\u05d9\u05ea \u05e2\u05d1\u05d5\u05e8 \u05d4\u05e7\u05d5\u05d1\u05e5 ''{0}''.
```
silentsakky: can we use " double quotes instead of ' single quote?

mihaipopahtp: We could use a smart quote (’ - \u2019) instead of doubling apostrophes. 
This will show up in strings with and without placeholders and there's not need to do anything else about it. 
I've done that for Catalan and it worked out fine.
I can do that for all languages if you agree to this.

zx-fulvio: @silentsakky Initially I went for double quotes since you can find double quotes in the same situation in the same file for different keys.
But the Hebrew localization is rather heavy on the use of apostrophes. Since I didn't whether this was a preference of the translator or rather a particular use of the character in Hebrew I stopped at quoting correctly the apostrophes.

zx-fulvio: @mihaipopahtp That would be my original idea as well. Problem is, U+2019 is a "directional" quote, something the apostrophe isn't (technically). As a quick fix it would be fine, but although it would be perfectly appropriate as a replacement for `'` in `isn't`, for instance, it might not be the case for something like `the account 'a@example.com'` where left and right quotation marks would perhaps be better. And that without considering non-European languages which likely have different requirements. But that would up to the translator to know.

mihaipopahtp: @zx-fulvio @silentsakky , 
In order to standardize this I will change the Catalan smart quotes to apostrophes, as the other languages use.

zx-fulvio: @mihaipopahtp Standardization is certainly a good thing. 
Just to clarify my previous point: the lines where I fixed the representation of apostrophes are just a portion of the total: merely the ones I could identify through the presence of format constructs.
Any other string used as a format which isn't one of those is still broken when it comes to apostrophes.
So the patch is a partial fix.
There are two or three options that I can see to fix the problem in its entirety:
(I) check string by string in the code to see whether they are literal strings or formats and write the apostrophes accordingly. **A lot** of work which would not prevent issues with strings added in the future or future modifications to existing string values. Also hard to check for errors in an automatic fashion.
(II) Get rid of all apostrophes everywhere, replacing them with a single character. The smart quote we both mentioned (U+2019) would be a decent candidate, because, even if it doesn't fit perfectly as I mentioned in the previous comment, it could be used where an apostrophe would be spot on and also as an enclosing character where one might use a double quote for instance.
(III) Get rid of all apostrophes using a human translator for each language to decide which character to use to replace them with _at every point_ in the strings. Obviously the best option, but also the most time-consuming and the one involving the most people.
Both (II) and (III) have the advantage of eliminating the problem at the root, also allowing to spot future problems just by checking for the presence of apostrophes in the text. Among the two, (II) which you suggested is most likely the best compromise. It's not perfect but should prevent glaring failures in the localized output from occurring again and it amounts to little more than a search and replace. IMHO an incorrect use of the quote would be preferable to a broken localized string.
My two (or four given the length ;-) ) cents.
Of course, any of the three options is an improvement over the current situation.

silentsakky: Thanks @zx-fulvio can you please update your PR to use smart quotes? for other languages @mihaipopahtp can handle it at global level

mihaipopahtp: @zx-fulvio

Totally agree with what you said above. 
I was going to get back to you on this one, because I recently pulled all the repos for some general checks I wanted to make and looked for this apostrophe issue as well.

My initial impression was that most apostrophes were doubled where needed, and Catalan was the one exception where this was treated differently. It's not and indeed there's a lot of fixing needed in all files/languages.

I think method (II) is the absolute best in this case.
For our other clients that use this file format, the smart quote is the preferred method and for the languages we have in common with them I know a smart quote is accepted. I will have to confirm this for languages like Hebrew, where this might be an issue, but I can fix this with either bidi marks or simply doubling the quotes if I have no other choice.
for instances such as 'a@example.com' I'm thinking of using opening and closing single quotes, if you agree.

I have a semi-automated method of doing this globally and it works decently. I've done it for Catalan and it went well.
I can do this in all languages (yours included and add you for review if you want).

silentsakky: sure no problem that should be fine

zx-fulvio: @mihaipopahtp No problem. I'd be happy to take a look and review the changes. Another set of eyes (and code) on it to spot discrepancies wouldn't hurt I think ;-) . 
@silentsakky I agree that tackling this specific issue on a global level should be a better solution than a piecemeal approach with a handful of pull requests.
Thanks to both for the feedback.

zx-fulvio: @mihaipopahtp Almost forgot. As an aside I've noticed how the Norwegian and Catalan localization files appear to be the only ones with CRLF line endings instead of LF line endings, at least in the few repositories I checked.
It's not a problem but I don't know if it might be the case to change that for the sake of consistency.
Just saying.

mihaipopahtp: I think CRLF line ending is a default for files saved under Windows.  All files that were created by me will have this (Vietnamese included) but I can fix it after I'm done checking the apostrophes.

zx-fulvio: You think correctly :) . Thanks for the bother.

PR #118, opened 2017-08-11
ZCS-1213: Universal UI - Update the attach vCard interface in the preferences

kalpeshhpatel:
#### Summary of Changes:
* Replaced input field of vCard widget with label.
* Updated template to group clear button and vCard
* Made attached vCard to look like address bubble.

https://jira.corp.synacor.com/browse/ZCS-1213


WebRoot/js/zimbraMail/mail/view/prefs/ZmSignaturesPage.js
```diff
@@ -773,7 +778,7 @@ function(skipControls, autoAdded) {
     signature._autoAdded = autoAdded;
 	signature = this._addSignature(signature, skipControls);
 	setTimeout(this._sigName.focus.bind(this._sigName), 100);
-
+	Dwt.setVisibility(this._vcardWrapper, false); //hide vCard container
```
tarangkhandelwal: when do we need to hide the container?

kalpeshhpatel: When no vCard is attached with signature, we need to hide it.

PR #119, opened 2017-08-14
ZCS-2348 XSS possible in webclient using malicious folder name

silentsakky:
Issue:
- User can do XSS attack using malicious name as folder name and searching for that folder

Resolution:
- Encode folder name before adding it to view of folder name suggestion dropdown in search box

**ZCS-2348 XSS possible in webclient using malicious account/location name**
Issue:
- User can do XSS attack using malicious name as account/location name and clicking on suggest time/location link in appointment edit view

Resolution:
- Encode text in ZmPopupMenu view, this will make sure XSS code can not be executed when showing popup menu

**ZCS-2348 XSS possible in webclient using malicious account name**
Issue:
- User can do XSS attack using malicious name as account name and adding that account as delegate from account preferences section

Resolution:
- Encode text when showing it as toast message
- Encode text in ZmRevokeShareDialog

### DIFF
--- a/WebRoot/js/zimbraMail/share/view/ZmTagMenu.js
+++ b/WebRoot/js/zimbraMail/share/view/ZmTagMenu.js
@@ -194,7 +194,7 @@ function(tagList, addRemove) {
 	var removeid = map ? (map + "_removetag"):this._htmlElId + "|REMOVETAG";
 
 	var miNew = this._menuItems[ZmTagMenu.MENU_ITEM_ADD_ID] = new DwtMenuItem({parent:this, id: addid});
-	miNew.setText(AjxStringUtil.htmlEncode(ZmMsg.newTag));
+	miNew.setText(ZmMsg.newTag);
 	miNew.setImage("NewTag");
 	miNew.setShortcut(appCtxt.getShortcutHint(this._keyMap, ZmKeyMap.NEW_TAG));
 	miNew.setData(ZmTagMenu.KEY_TAG_EVENT, ZmEvent.E_CREATE);
@@ -204,7 +204,7 @@ function(tagList, addRemove) {
 	// add static "Remove Tag" menu item
 	var miRemove = this._menuItems[ZmTagMenu.MENU_ITEM_REM_ID] = new DwtMenuItem({parent:this, id: removeid});
 	miRemove.setEnabled(false);
-	miRemove.setText(AjxStringUtil.htmlEncode(ZmMsg.removeTag));
+	miRemove.setText(ZmMsg.removeTag);
 	miRemove.setImage("DeleteTag");
 
 	var removeList = addRemove.remove;
--- a/WebRoot/js/zimbraMail/share/view/dialog/ZmRevokeShareDialog.js
+++ b/WebRoot/js/zimbraMail/share/view/dialog/ZmRevokeShareDialog.js
@@ -73,7 +73,7 @@ function(share) {
 	var isAllShare = share.grantee && (share.grantee.type == ZmShare.TYPE_ALL);
 
 	var params = isPubShare ? ZmMsg.shareWithPublic : isGuestShare ? share.grantee.id : isAllShare ? ZmMsg.shareWithAll :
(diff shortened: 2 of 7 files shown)


WebRoot/js/zimbraMail/share/view/ZmTagMenu.js
```diff
@@ -194,7 +194,7 @@ function(tagList, addRemove) {
 	var removeid = map ? (map + "_removetag"):this._htmlElId + "|REMOVETAG";
 
 	var miNew = this._menuItems[ZmTagMenu.MENU_ITEM_ADD_ID] = new DwtMenuItem({parent:this, id: addid});
-	miNew.setText(AjxStringUtil.htmlEncode(ZmMsg.newTag));
```
kkakani: Why do we need to encode a string from ZmMsg?

WebRoot/js/zimbraMail/share/view/dialog/ZmRevokeShareDialog.js
```diff
@@ -74,7 +74,7 @@ function(share) {
 
 	var params = isPubShare ? ZmMsg.shareWithPublic : isGuestShare ? share.grantee.id : isAllShare ? ZmMsg.shareWithAll :
 					(share.grantee.name || ZmMsg.userUnknown);
-	this._confirmMsgEl.innerHTML = this._formatter.format(params);
+	this._confirmMsgEl.innerHTML = this._formatter.format(AjxStringUtil.htmlEncode(params));
```
kkakani: Why do we need to encode params? Shouldn't we only be encoding the share.grantee.id and/or share.grantee.name instead?

kkakani: Why encode a string we are fetching from ZmMsg.properties. Is there any way it can be malicious?

WebRoot/js/zimbraMail/share/view/ZmTagMenu.js
silentsakky: usually it is encoded only when we have a placeholder in the message string
but not sure why it was needed here
this is not needed anymore as we have moved encoding to ZmPopupMenu.js

WebRoot/js/zimbraMail/share/view/dialog/ZmRevokeShareDialog.js
silentsakky: encoding is only needed when we have a placeholder in the string
to remove confusion I have moved encoding to only data part

WebRoot/js/zimbraMail/share/view/dialog/ZmRevokeShareDialog.js
```diff
@@ -72,8 +72,8 @@ function(share) {
 	var isGuestShare = share.isGuest();
 	var isAllShare = share.grantee && (share.grantee.type == ZmShare.TYPE_ALL);
 
-	var params = isPubShare ? ZmMsg.shareWithPublic : isGuestShare ? share.grantee.id : isAllShare ? ZmMsg.shareWithAll :
-					(share.grantee.name || ZmMsg.userUnknown);
+	var params = isPubShare ? ZmMsg.shareWithPublic : isGuestShare ? AjxStringUtil.htmlEncode(share.grantee.id) : isAllShare ? ZmMsg.shareWithAll :
+					(AjxStringUtil.htmlEncode(share.grantee.name) || ZmMsg.userUnknown);
 	this._confirmMsgEl.innerHTML = this._formatter.format(params);
```
tarangkhandelwal: For me, the fix was good to be at line 77, just because this is the moment we are manipulating a DOM where XSS attacks actually happen.

tarangkhandelwal: As discussed offline, this is fine to go in this case.

silentsakky: changed this to only encode grantee name and not id

PR #122, opened 2017-08-17
ZCS-1217:- Universal UI - Change password visual update

SatyamKumarOnline:
Desc:- Changes to accommodate the change password related layout.
Added placeholder for respective fields.
added message key/value  for Old Password label.

Changeset :-
1. WebRoot/css/login.css
2. WebRoot/h/changepass
3. WebRoot/messages/ZmMsg.properties
4. WebRoot/skins/_base/base4/skin.properties


WebRoot/css/login.css
```diff
@@ -129,6 +129,29 @@ FORM {
 	@cssShadow@:0px 0px 2px black;
 }
 
+.ZChangePassLogin {
```
tarangkhandelwal: This property styles are getting overridden:
https://screencast.com/t/2a9SEl4x

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -101,8 +101,13 @@ LoginBannerPosition =   background-size: contain;
...
 LoginCopyrightColor =   color: @hexToRGBA(AppC, 60)@;
 LoginInputField     =   width: 100%;  height: 40px;  @MediumRoundCorners@;  border: 1px solid @hexToRGBA(SecondaryBrandingColor, 50)@;  padding: 6px 10px;  @FontSize-big@
-
-InlineErrorPanel    =   color: @AppC@;  @FontSize-normal@
+ChangePasswordLoginSection        = top:40%;
+ChangePassWordLoginInfoPanel      = margin-bottom:@SkinWrapperPadding@;
+ChangePassWordLoginInfoPanelCell  = color:@AppC@; @FontSize-slightly-big@
+ChangePasswordLoginConfirmCell    = padding-bottom:0px;
+ChangePassWordLoginErrorMessage   = @ChangePassWordLoginInfoPanelCell@; padding-top:@SkinWrapperPadding@; padding-bottom:0px;
+ChangePasswordButton              = padding-top:30px;
```
tarangkhandelwal: why have a padding for the button ? it would be generic if all the ". zLoginFieldContainer" elements including change password field has the padding-bottom. 
And then we won't need the line 107 here setting padding 0.

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -101,8 +101,13 @@ LoginBannerPosition =   background-size: contain;
 LoginFooterColor    =   color: @AppC@;
 LoginCopyrightColor =   color: @hexToRGBA(AppC, 60)@;
 LoginInputField     =   width: 100%;  height: 40px;  @MediumRoundCorners@;  border: 1px solid @hexToRGBA(SecondaryBrandingColor, 50)@;  padding: 6px 10px;  @FontSize-big@
-
-InlineErrorPanel    =   color: @AppC@;  @FontSize-normal@
+ChangePasswordLoginSection        = top:40%;
+ChangePassWordLoginInfoPanel      = margin-bottom:@SkinWrapperPadding@;
+ChangePassWordLoginInfoPanelCell  = color:@AppC@; @FontSize-slightly-big@
+ChangePasswordLoginConfirmCell    = padding-bottom:0px;
+ChangePassWordLoginErrorMessage   = @ChangePassWordLoginInfoPanelCell@; padding-top:@SkinWrapperPadding@; padding-bottom:0px;
```
tarangkhandelwal: padding top may not be required if we have padding bottom to the confirm password field similar to other fields.

WebRoot/css/login.css
```diff
@@ -129,6 +129,29 @@ FORM {
...
+
+.ZChangePassLogin .errorText {
+	@ChangePassWordLoginErrorMessage@
+}
+
+.ZChangePassLogin .ZChangePassConfirm {
+	@ChangePasswordLoginConfirmCell@
+}
+
+.ZChangePassLogin .ZChangePassButton {
```
tarangkhandelwal: the button should also have cursor:pointer.

tarangkhandelwal:
@SatyamKumarOnline : I don't think the container in the change password screen is really in center. 
I would suggest to have a look at: 
https://stackoverflow.com/questions/396145/how-to-vertically-center-a-div-for-all-browsers
http://emergentweb.com/test/valign.html

WebRoot/h/changepass
```diff
@@ -81,53 +77,53 @@
...
+                                    <td class='errorText'>
+                                        <c:out value="${errorMessage}"/>
+                                    </td>
+                                </tr>
+                            </c:if>
+
                             <tr>
-                                <td></td>
-                                <td><input type=submit class='zLoginButton'
+                                <td class='ZChangePassButton'><input type=submit class='zLoginButton'
```
tarangkhandelwal: why do we need ZChangePassButton class? the styling should really be same as of zLoginButton in my opinion.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: It's a padding for the button container.
We need a padding- top of 16 px for the error message.
So, we over-rided generic  padding -bottom for  confirm cell.

WebRoot/skins/_base/base4/skin.properties
SatyamKumarOnline: Reply above.

WebRoot/h/changepass
SatyamKumarOnline: As discussed above, we have overriden the padding-top for the change password button.
So, needing a class here for reference.

WebRoot/skins/_base/base4/skin.properties
tarangkhandelwal: what I see from the mockups is the space between inputs fields & space between confirm input field & button is same. I don't think we need ChangePasswordLoginConfirmCell setting padding-bottom to 0.

kkakani:
@SatyamKumarOnline The codacy check has failed. Are you going to look into it?

WebRoot/h/changepass
tarangkhandelwal: it looks odd as in structure to have an element with class "ZChangePassButton" and its child with class "zLoginButton". Do you want to get rid of class and use pseudo selector? :last-child maybe?

SatyamKumarOnline:
@kkakani , the code check is failing because I believe we are having our own custom defined property in properties file with start and ending like @@, which is not browser will recognise while parsing CSS.

WebRoot/h/changepass
SatyamKumarOnline: In some my previous PR, I was suggested not to use last-child, so avoided doing that.
The reason being if the skin develop changes templates, it may affect the layout.
So, not using that.
@silentsakky , what would you say, here.

kalpeshhpatel: Will `. ZChangePassLogin  input[type='submit']`  works?

SatyamKumarOnline: Seems, it works.  That good to go.

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -84,11 +84,12 @@ DwtImgDir           =   @AppContextPath@/img/dwt
...
 LoginBkgdC2         =   @darken(AppC,7)@
 LoginBkgdGrad       =   
-LoginScreen         =   background-color: @AltC@; @LoginBkgdGrad@
+LoginScreen         =   background-color:@AltC@; @LoginBkgdGrad@; @FontFamily-default@; @FontSize-normal@; @FullWidth@; @FullHeight@; position:absolute; display:table; overflow:hidden;
+LoginScreenCenter   =   overflow:visible; display:table-cell; vertical-align:middle; z-index:11;
 LoginContentC1      =   @AltC@
 LoginContentC2      =   transparent
 LoginContentGrad    =   
-LoginContentStyle   =   width: 270px; min-height: 270px;
+LoginContentStyle   =   width: 270px; min-height: 270px; margin-left:auto; margin-right:auto;
```
tarangkhandelwal: can this be margin:0 auto?

PR #123, opened 2017-08-18
ZCS-1215: Universal UI - Filters in preferences, visual updates

kalpeshhpatel:
### Summary of Changes:
* Major changes are, in filter preference page, we have replaced `DwtChooser` widget with simple `DwtListView` and accordingly we have removed some of the no-longer needed class and added some new classes.
* ZmFilterRuleController.js
  * Updated various method to remove many conditions related to two different list view, one for active filers and other for available filters.
  * Now both active filters and available filters are combined into single list view and accordingly code is updated
  * Renamed some of the event listeners method for consistency purpose.
  * Added new event listener `ZmFilterRulesController.prototype._toggleStateListener` to change state of filter rule.
  * Removed filter list selection listener from this class as it will be handled in `ZmFilterRulesView` class now.
* ZmFilterRulesView.js:
  * At high level, removed classes `ZmFilterRulesChooser`, `ZmFilterChooserActiveListView` and `ZmFilterChooserNotActiveListView` and it’s related methods, since it is not needed now.
  * Added new class `ZmFilterListView` which extends `DwtListView` to handle filter rules list.
  * Implemented required methods for `ZmFilterListView ` class.
* Updated template for filter page as per requirement and also removed unused template
* Updated icons for filter related operations
* Added new svg icons as per requirement
* Added `ZAltButton` style for toolbar button as well
* Globally change active tab style as per mockup

https://jira.corp.synacor.com/browse/ZCS-1215


WebRoot/js/zimbraMail/prefs/controller/ZmFilterRulesController.js
```diff
@@ -499,6 +459,23 @@ function(ev) {
 };
 
 /**
+ * "Active" rule check-box is checked or unchecked
+ * 
+ * @ev		[DwtEvent]		the click event
```
tarangkhandelwal: I think, this should be :
@param	{DwtEvent}	ev		the click event

kalpeshhpatel: Ohh yes, my bad. I'll correct it.

Thanks for highlighting

kalpeshhpatel: Corrected here: https://github.com/Zimbra/zm-web-client/pull/123/commits/d8d573a07657e66d436b51f5397758620d661f22#diff-9d2d8925c8355c1c28606cf3376628a3R464

WebRoot/js/zimbraMail/prefs/view/ZmFilterPage.js
```diff
@@ -36,6 +36,7 @@
  */
 ZmFilterPage = function(parent, section, controller) {
 	ZmPreferencesPage.apply(this, arguments);
+	this.addClassName("ZmFilterPage");
```
silentsakky: better to pass this as className in params before calling parent contructor

WebRoot/js/zimbraMail/prefs/view/ZmFilterPage.js
```diff
@@ -50,6 +51,7 @@ function() {
 		this._activityStreamsButton = new DwtButton({parent:this, parentElement: this._htmlElId+"_ACTIVITY_STREAM_BUTTON" });
 		this._activityStreamsButton.setText(ZmMsg.activityStreamSettings);
 		this._activityStreamsButton.addSelectionListener(new AjxListener(this, this._activityStreamDialog));
+		this._activityStreamsButton.addClassName("ZInlineButton");
```
silentsakky: same as above

WebRoot/js/zimbraMail/prefs/view/ZmFilterRulesView.js
```diff
@@ -35,18 +35,15 @@
  */
 ZmFilterRulesView = function(parent, controller) {
 
-	DwtTabViewPage.call(this, parent, "ZmPreferencesPage ZmFilterRulesView", Dwt.STATIC_STYLE);
+	DwtTabViewPage.call(this, parent, "ZmFilterRulesView", Dwt.STATIC_STYLE);
 
 	this._controller = controller;
 	this._prefsController = AjxDispatcher.run("GetPrefController");
 
-	this._rules = AjxDispatcher.run(controller.isOutgoing() ? "GetOutgoingFilterRules" : "GetFilterRules");
```
silentsakky: any reason to remove this?

kalpeshhpatel: This change listener has been moved to https://github.com/Zimbra/zm-web-client/pull/123/files/d8d573a07657e66d436b51f5397758620d661f22#diff-ada83781749503b4db7ab2bcf7361113R190

It was there initially but not needed now.

WebRoot/js/zimbraMail/prefs/view/ZmFilterRulesView.js
```diff
@@ -79,20 +76,17 @@ function() {
 											 context:ZmId.VIEW_FILTER_RULES});
 		this._toolbar.replaceElement(toolbarEl);
 		this._tabGroup.addMember(this._toolbar);
+		//add `ZAltButton` style to 'Create Filter' toolbar button
+		var addFilterBtn = this._toolbar.getButton(ZmOperation.ADD_FILTER_RULE);
+		addFilterBtn && addFilterBtn.delClassName('ZToolbarButton','ZButton ZAltButton'); //to look-it like ZButton
```
silentsakky: can we do this when creating a button, that way we don't have to remove existing class and add new

kalpeshhpatel: I tried it initially, but in this case it was not working. Since we are creating ZmToolbar here and it takes care of creating buttons directly returned by `ZmFilterRulesController.prototype.getToolbarButtons`. This method just returns `ZmOperation` instance and not instance of `DwtButton` or `DwtControl`. 

Hence needed to follow this approach.

Thanks

WebRoot/skins/_base/base4/skin.properties
```diff
@@ -2434,3 +2434,22 @@ PrefCalWorkHoursCell    = padding:0 @PrefCalInternalSpacing@;
...
+# Preference- Filters interface
+######################################
+
+PrefFilterPage                  = 
+PrefFilterTabView               = margin:@PrefViewInnerSpacing@ 0 0 0;
+PrefFilterTabBar                = @PanelBorder@; @MediumRoundCorners@; @blRadius@:0px; @brRadius@:0px;
+PrefFilterViewToolbar           = margin:@PrefViewWrapperPadding@ 0 0 0;
+PrefFiltersList                 = @PanelBorder@; border-top:none; @MediumRoundCorners@; @tlRadius@:0px; @trRadius@:0px;
+PrefFiltersListEmpty            = padding:@SkinWrapperPadding@;
+PrefFilterCellLayout            = display:flex; justify-content:space-between; align-items:center;
```
silentsakky: this looks problamatic for IE11, https://caniuse.com/#search=justify-content
can you check?

kalpeshhpatel: Maybe we should revisit this after testing in IE11, as is seems to have partial support and must support basic flex properties at-least. (And these are very basic flex properties we are using) And if it's working fine, no issues then. 

It works perfectly for IE Edge, and I have tested it.

What do you suggest?

Thanks

WebRoot/js/zimbraMail/prefs/view/ZmFilterRulesView.js
silentsakky: ok

WebRoot/js/zimbraMail/prefs/view/ZmFilterPage.js
kalpeshhpatel: Well, parent constructor for now doesn't seems to support className param for now. (https://github.com/Zimbra/zm-web-client/blob/develop/WebRoot/js/zimbraMail/prefs/view/ZmPreferencesPage.js),  we will need to update `ZmPreferencesPage` class to have support for `className` param.

Do you feel we should go ahead and have this changes in `ZmPreferencesPage`?

Let me know, thanks.

WebRoot/js/zimbraMail/prefs/view/ZmFilterPage.js
kalpeshhpatel: I'll update this and let you know.

kalpeshhpatel: Updated here: https://github.com/Zimbra/zm-web-client/pull/123/commits/250176b7d7195a1cf3ddd2a670199fd5dc7d702b#diff-fc2845c8c632ff84b21927f1ef3d5789L51

WebRoot/skins/_base/base4/skin.properties
silentsakky: it should not break drastically in IE 11, some small minor issues can be ignored for now and fixed later

WebRoot/js/zimbraMail/prefs/view/ZmFilterPage.js
silentsakky: it's ok I thought we are supporting passing any number of arguments as part of params
ignore this comment

WebRoot/skins/_base/base4/skin.properties
kalpeshhpatel: Yes, agree, it should not.

PR #124, opened 2017-08-18
ZCS-1294:  Compose view text inputs should align with the text editor

tarangkhandelwal:
ZCS-1294  Compose view text inputs should align with the text editor in all cases.

Issue: The empty padding td at the end is taking the available space which is more in case where from drop down is not available.

Fix: Adding width as 1% to the last div that’s there for space so that it doesn’t take all available space.

Changeset:
* skin.properties:  Adding width of 1% to ComposeViewHeaderEmptyRight property variable.

### DIFF
--- a/WebRoot/skins/_base/base4/skin.properties
+++ b/WebRoot/skins/_base/base4/skin.properties
@@ -1311,7 +1311,7 @@ ComposeViewSubjectInputField= border:none; color:@TxtC@; height:3rem; @FullWidth
 ComposeViewTinyMceContainer =
 ComposeViewHtmlEditor       = padding:@ComposeViewOutsidePadding@; background-color: @AppC@;
 ComposeViewPlainTextEditor  = @FontSize-big@ @FontFamily-fixed@ @FullWidth@ @AppBorder@ overflow:auto; padding:8px;
-ComposeViewHeaderEmptyRight = padding-right: 18px; width: 1%;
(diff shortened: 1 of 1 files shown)


WebRoot/skins/_base/base4/skin.properties
```diff
@@ -1311,7 +1311,7 @@ ComposeViewSubjectInputField= border:none; color:@TxtC@; height:3rem; @FullWidth
 ComposeViewTinyMceContainer =
 ComposeViewHtmlEditor       = padding:@ComposeViewOutsidePadding@; background-color: @AppC@;
 ComposeViewPlainTextEditor  = @FontSize-big@ @FontFamily-fixed@ @FullWidth@ @AppBorder@ overflow:auto; padding:8px;
-ComposeViewHeaderEmptyRight = padding-right: 18px;
+ComposeViewHeaderEmptyRight = padding-right: 18px; width: 1%;
```
silentsakky: not sure but empty space would be reserved for scroll bar area, can we change it to px or em?

tarangkhandelwal: 1px works too . I can change to that.

tarangkhandelwal: fixed here:
5783c3eaafad2721839a6206d8ec4b7722e40ba4

PR #129, opened 2017-08-24
ZCS-2449: Universal UI - Folder Properties Dialog

tarangkhandelwal:
Description: Fixing alignment issues for the inline buttons in the dialog.

Changeset:
 * zm.css: Matching font size for button text with the dialog content.
 * skin.css: Aligning inline button in dialog with the dialog content.

### DIFF
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -6669,7 +6669,7 @@ a.ZmLinkDisabled {
 
 .ZDialogButton .ZButtonTable .ZWidgetTitle {
     color: @AltC@;
-    letter-spacing: 1px;
+    @FontSize-slightly-big@
 }
 
 /* Primary button style */ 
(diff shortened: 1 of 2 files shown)


WebRoot/css/zm.css
```diff
@@ -6669,7 +6669,7 @@ a.ZmLinkDisabled {
 
 .ZDialogButton .ZButtonTable .ZWidgetTitle {
     color: @AltC@;
-    letter-spacing: 1px;
+    font-size:1.1rem;
```
silentsakky: @FontSize-slightly-big@

tarangkhandelwal: changing it to use the property. Not moving these styles to a property variables as I think there would be more efforts moving all dialog related stylings into variable types.

tarangkhandelwal: Fixed here:
3f69c82ea0a80ec5e23f8f79e037f97225ba4c43

PR #131, opened 2017-08-30
ZCS-2415 Universal UI: Add New Calendar dialog

silentsakky:
- make name field take full width
- fix right padding for select list
- fix checkbox positioning for feed field

### DIFF
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -4366,6 +4366,14 @@ HTML>BODY .appt-selected .appt_allday_body
 	@ChooseFolderDialogButtonTitle@
 }
 
+.ZmCreateNewFolderDialog .NameField {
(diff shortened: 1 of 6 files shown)


WebRoot/css/zm.css
```diff
@@ -4372,6 +4372,14 @@ HTML>BODY .appt-selected .appt_allday_body
 	@ChooseFolderDialogButtonTitle@
 }
 
+.CreateNewFolderDialog .NameField {
+	@CreateNewFolderDialogNameField@
+}
+
+.CreateNewFolderDialog .ColorLabel {
```
tarangkhandelwal: can we add prefix "Zm" in class names?

silentsakky: Fixed

PR #139, opened 2017-09-08
ZCS-2412: Universal UI- Calendar Appointment tooltip

kalpeshhpatel:
Summary of Changes:
* Updated look and feel of calendar appointment tooltip as per new mockup

https://jira.corp.synacor.com/browse/ZCS-2412

### DIFF
--- a/WebRoot/css/zm.css
+++ b/WebRoot/css/zm.css
@@ -2866,13 +2866,6 @@ td.calendar_month_day_item table td {
 	@MiniCalTooltip-description@
 }
 
-.calendar_tooltip_attendees {
-	border-bottom:1px solid black;
-	padding:1px;
-	white-space:nowrap;
-	font-weight:bold;
-}
-
 .suggestion_tooltip_attendee {
 	padding:3px;
 }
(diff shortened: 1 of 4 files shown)


WebRoot/css/zm.css
```diff
@@ -7597,4 +7590,33 @@ input[type="radio"] + label {
...
+
+.ZmAppointmentTooltip * {
+	@FontSize-slightly-big@
+}
+
+.ZmAppointmentTooltip h3 {
+	@AppointmentTooltipTitle@
+}
+
+.ZmAppointmentTooltip h6 {
```
tarangkhandelwal: Suggestion, should we be replacing heading elements with div/span and apply sizes using css . Not sure what general practice is ?

kalpeshhpatel: Before it was using div, span for showing title. I have changed it to heading tag because it represents title and hence more suitable here.

Thanks

tarangkhandelwal: Thats fine. Just one thing that I wasn't sure was why are we overriding font size for heading tags? They themselves are meant to have different sizes right?

kalpeshhpatel: Heading tags are not meant to have different size. They meant to have different level of importance semantically. We can style them anyway we want. Let me know if you still have any question.

tarangkhandelwal: thats fine then. Thanks

PR #142, opened 2017-09-13
Fixing: Free Busy calendar view breaking with latest changes.

tarangkhandelwal:
Fixing: Free Busy calendar view breaking with latest changes.

Root cause: The same view is used in new appointment where the start date is displayed inside the scheduler. In case of Free Busy view, there won’t be any appointment associated and thus this._appt is undefined.

Changeset:
* Adding condition to check for presence of this._appt to be defined.
* Adding ZmOperation.FB_VIEW to the list of tabs to be styled as tab in Calendar App.
* Changing the delete action button to use ImgCancel.svg

### DIFF
--- a/WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
+++ b/WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
@@ -1072,7 +1072,8 @@ function(buttons) {
 		ZmOperation.WORK_WEEK_VIEW,
 		ZmOperation.WEEK_VIEW,
 		ZmOperation.MONTH_VIEW,
-		ZmOperation.CAL_LIST_VIEW];
+		ZmOperation.CAL_LIST_VIEW,
+		ZmOperation.FB_VIEW];
 
 	var overrides = {};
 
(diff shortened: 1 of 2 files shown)


WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
```diff
@@ -1072,7 +1072,8 @@ function(buttons) {
 		ZmOperation.WORK_WEEK_VIEW,
 		ZmOperation.WEEK_VIEW,
 		ZmOperation.MONTH_VIEW,
-		ZmOperation.CAL_LIST_VIEW];
+		ZmOperation.CAL_LIST_VIEW,
+		ZmOperation.FB_VIEW];
```
kalpeshhpatel: Will this make sure to show Free/Busy link in toolbar only when it is configured?

tarangkhandelwal: Yes.  ZmCalViewController constructor function takes care of that. 
https://screencast.com/t/ANNUtWnajmPI

PR #148, opened 2017-09-20
When the password is changed, the success message no show.

ljshj:


WebRoot/h/changepass
```diff
@@ -73,7 +73,8 @@
     <fmt:message key="favIconUrl" var="favIconUrl"/>
     <link rel="SHORTCUT ICON" href="<c:url value='${favIconUrl}'/>">
 </head>
-<body <c:if test="${successfullLogin ne 'true'}">
+<body <c:if test="${
```
silentsakky: are we missing something here?

silentsakky: no response, so closing this PR

PR #149, opened 2017-09-21
Updated schema Version

sbluhm:
The tag async-supported only exists from schema version 3.0 onwards. So the XML version was upped and both schema location changed to the correct ones


silentsakky:
Merged in https://github.com/Zimbra/zm-web-client/commit/72128088588ce6e64961f0cbc5221c44e6665398

PR #155, opened 2017-09-27
ZCS-2972 redirect based on domain's public service hostname

tmclane:
The login.jsp was updated to call the new `zm:buildRedirectURL` defined in `zm-taglib` in https://github.com/Zimbra/zm-taglib/pull/4

* c:redirect - updated to redirect the request to the absolute URI returned by `zm:buildRedirectURL`.
This prevents the `Host` header from redirecting the web browser to a hostile domain intercepting the ZM_AUTH_TOKEN cookie.

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -188,7 +188,7 @@
                         <c:choose>
                             <c:when test="${client eq 'socialfox'}">
                                     <c:set var="sbURL" value="/public/launchSidebar.jsp"/>
-                                    <c:redirect url="${sbURL}">
+                                    <c:redirect url="${zm:buildRedirectURL(pageContext, sbURL)}">
                                         <c:forEach var="p" items="${paramValues}">
                                             <c:forEach var='value' items='${p.value}'>
                                                 <c:set var="testKey" value=",${p.key},"/>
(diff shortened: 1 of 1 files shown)


tmclane:
Fully tested on AWS using Split-UI configuration.  Works fine with the caveat after the successful login a redirect occurs which causes a second login to be required.  This only happens if the original URL is not the same as what is configured as the `zimbraPublicServiceHostname` value.

mss:
I already left a longer comment on zimbra/zm-nginx-conf#6 but here are two use cases which probably won't work anymore with this change:


First, from [BZ-108299](https://bugzilla.zimbra.com/show_bug.cgi?id=108299):

- The actual hostname is `mail.example.local` which resolves to a RFC 1918 IP
- The server is available publically via NAT on `mail.example.com`
- `zimbraPublicServiceHostname` is not set

Result with 8.7: While this config is suboptimal, a login tended to work out of the box.
Expected result with this change: Login will fail for users using the public host name since the redirect will point to the internal name.

Suggested fix:
* Add this to the admin docs/install guide
* Make the installer ask for the `zimbraPublicServiceHostname` 

Second:

- The `zimbraPublicServiceHostname` is set to `mail.example.com`
- The server is also available via `mail.example.org`

Result with 8.7: You could visit either hostname and log in and stay on the hostname.
Expected result with this change: Login on `mail.example.org` will seemingly succeed but the user is redirected to `mail.example.com` and the user will have to login again since the Cookie is not set

Suggested fix:
* Check not only against `zimbraPublicServiceHostname` and the server name but also all `zimbraVirtualHostname`s.
* Add this to the admin docs/install guide

tmclane:
I am looking into adding the other information into the redirect / no-redirect algorithm at this time.

tmclane:
Based on discussions surrounding the behavior in general I am closing this without merging for the time being.

PR #157, opened 2017-09-28
ZCS-2948: add temporary jsp for address verification response

shrikant-prasad:
when address verification link is clicked, response has to be redirected to some page, added this temporary jsp for this purpose. final jsp will be created by web client team under separate ticket.


silentsakky:
From ticket this looks like feature for nextgen client, do we need this feature in current webclient?

shrikant-prasad:
if someone enables this feature, then we should have the web client support as well because from server side it doesn't know which client it's interacting with.

PR #160, opened 2017-10-03
ZCS-2997: Universal UI - Reminder dialog should not change size for appointment with longer subjects

kalpeshhpatel:
Fix:
* Set `max-width` of reminder dialog and it’s subject as well
* Added custom css rule to handle very long calendar name in calendar select dropdown.

https://jira.corp.synacor.com/browse/ZCS-2997

### DIFF
--- a/WebRoot/skins/_base/base4/skin.css
+++ b/WebRoot/skins/_base/base4/skin.css
@@ -1277,3 +1277,8 @@ div[name='_mergeBtnContainerId'] .ZCalToggleBtnSelected .ZWidgetTitle {
 .ZmCalPrintDialog td[id$="todayButtonContainer"] .ZWidgetTitle {
     @AlignLeft@
 }
+
+.ZmApptQuickAddDialog .ZSelect .ZWidgetTitle {
+    max-width: 1px;
+    @OverflowEllipsis@
+}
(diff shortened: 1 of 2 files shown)


WebRoot/skins/_base/base4/skin.css
```diff
@@ -1276,4 +1276,9 @@ div[name='_mergeBtnContainerId'] .ZCalToggleBtnSelected .ZWidgetTitle {
 /* Today Inline button for Calendar Print Dialog */
 .ZmCalPrintDialog td[id$="todayButtonContainer"] .ZWidgetTitle {
     @AlignLeft@
-}
\ No newline at end of file
+}
+
+.ZmApptQuickAddDialog .ZSelect .ZWidgetTitle {
+    max-width: 1px;
```
silentsakky: having a max width of 1px, isn't this weird?

kalpeshhpatel: Yes, it sounds so. But it's actually combination of `width:100%` and `max-width:1px` which effectively helps us to achieve `text-overflow: ellipsis` behaviour here. And it works for all major browsers. 

You can find more here: https://stackoverflow.com/questions/9789723/css-text-overflow-in-a-table-cell

Thanks

PR #164, opened 2017-10-06
ZCS-2432: Universal UI - Create New Folder Dialog

kalpeshhpatel:
Updated look and feel of Create New Folder dialog as per mockup.

Changes:
* Added/updated required styles in `dwt.css`, `zm.css` and `skin.css`
* Added/updated required properties in `skin.properties`
* Added/updated templates as per requirement in `Dialogs.templates` and `Widgets.template`
* Added/updated required classes for various controls in `ZmDialog.js`, `ZmNewOrganizerDialog.js` and `ZmNewTagDialog.js`

https://jira.corp.synacor.com/browse/ZCS-2432

### DIFF
--- a/WebRoot/templates/share/Dialogs.template
+++ b/WebRoot/templates/share/Dialogs.template
@@ -400,21 +400,33 @@
 </template>
 
 <template id="share.Dialogs#ZmNewTagDialog">
-	<table role="presentation" class='ZPropertySheet' cellspacing='6'>
+	<table role="presentation" class='ZmDialogPropertySheet'>
 		<tr>
-			<td class='Label'><$= ZmMsg.tagName $>:</td>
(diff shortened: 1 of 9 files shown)


WebRoot/templates/share/Dialogs.template
```diff
@@ -400,21 +400,33 @@
 </template>
 
 <template id="share.Dialogs#ZmNewTagDialog">
-	<table role="presentation" class='ZPropertySheet' cellspacing='6'>
+	<table role="presentation" class='ZmDialogPropertySheet' cellspacing='6'>
```
silentsakky: do we need cellspacing here?

kalpeshhpatel: No, it's not needed. I'll remove it. Thanks for pointing it out.

Thanks

PR #169, opened 2017-10-11
ZCS-2716 Universal UI : Under Preferences->Sharing, Share dropdown shows 'briefcase' even if briefcase feature is disabled

silentsakky:
- Before creating select options check if feature is disabled or not


kalpeshhpatel:
Is this only issue for universal-ui? Is it working for existing web client?

silentsakky:
I haven't checked but issue should be present in current webclient also, should we fix this in develop then?

PR #173, opened 2017-10-12
ZCS-2995 Universal UI : Right Navbar options got disappeared on doing zoom out and then rest to default

silentsakky:
- When zooming out and resetting to default size calculation done by ZmToolbar.js was setting incorrect width for ZmAppChooser, so override method to not do any width calculation as ZmAppChooser will take available width automatically
- When zooming in/out ZmAppChooser was not adjusting it's button visibility based on available space, so fixed it


silentsakky:
Merged in https://github.com/Zimbra/zm-web-client/commit/4464a472a1c029ac691359bb8e461c94b16c249c

PR #175, opened 2017-10-12
UX Feedback

silentsakky:


WebRoot/js/zimbraMail/mail/view/ZmConvView2.js
```diff
@@ -1203,7 +1207,7 @@ ZmConvReplyView.prototype._initializeToolbar = function() {
 			buttons:			buttons,
 			overrides:			overrides,
 			posStyle:			DwtControl.STATIC_STYLE,
-			buttonClassName:	"DwtToolbarButton",
+			buttonClassName:	"ZToolbarButton",
```
kalpeshhpatel: Minor, if we can make indentation of couple of lines here consistent, it will be better.

Thanks

silentsakky: Tab is correct in editor

kalpeshhpatel: No problem. Looks good in my editor too.

PR #182, opened 2017-10-17
ZCS-2424: Universal UI- Color picker widget

kalpeshhpatel:
Updated look and feel of color picker widget as per mockup

https://jira.corp.synacor.com/browse/ZCS-2424

### DIFF
--- a/WebRoot/templates/dwt/Widgets.template
+++ b/WebRoot/templates/dwt/Widgets.template
@@ -609,14 +609,10 @@ NOTE: The outer table/tr is a necessary hack. It is NOT inserted in the
 </template>
 
 <template id='dwt.Widgets#DwtColorPickerButton' class='ZWidget'>
-	<table role="presentation" class='ZWidgetTable ZMenuItemTable ZMenuItemBorder'>
+	<table role="presentation" class='ZWidgetTable'>
 		<tr>
-			<td><div id='${id}_color' class='Color'></div></td>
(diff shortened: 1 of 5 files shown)


WebRoot/templates/dwt/Widgets.template
```diff
@@ -609,14 +609,10 @@ NOTE: The outer table/tr is a necessary hack. It is NOT inserted in the
...
-	<table role="presentation" class='ZWidgetTable ZMenuItemTable ZMenuItemBorder'>
+	<table role="presentation" class='ZWidgetTable'>
 		<tr>
-			<td><div id='${id}_color' class='Color'></div></td>
-			<td id='${id}_title' class='ZWidgetTitle'
-			<$ if (AjxEnv.isIE) { $>
-				style='width:auto;'
-			<$ } $>
-			></td>
+			<td><div id='${id}_color' class='Color' style="margin-right:4px;"></div></td>
```
silentsakky: should we remove margin-right? it is already getting applied through ColorPickerColor property

kalpeshhpatel: Here I need to intentionally override styles being applied here: https://github.com/Zimbra/zm-web-client/pull/182/files#diff-350ba35347724bc4095d065d05880265R907 and https://github.com/Zimbra/zm-web-client/pull/182/files#diff-350ba35347724bc4095d065d05880265R911

Hence added inline style.

Thanks

PR #187, opened 2017-11-02
ZCS-3492 Universal UI: Schedule shows time slot at the wrong position

silentsakky:
- correct heading of scheduler headers as we are starting time from 12PM/0 instead of 1AM/1


silentsakky:
Merged in https://github.com/Zimbra/zm-web-client/commit/9fba6b3928048514a8def67f0503a947fb56d3b8

PR #206, opened 2017-11-15
ZCS-3233: Universal UI-  Match topography for mail list views

kalpeshhpatel:
Updated typography of mail list view as per mockup.

https://jira.corp.synacor.com/browse/ZCS-3233

### DIFF
--- a/WebRoot/skins/_base/base4/skin.css
+++ b/WebRoot/skins/_base/base4/skin.css
@@ -653,7 +653,7 @@ SELECT                      {   @NativeSelect@  }
 /*Temporary overrides that need to move to skin properties*/
 .RowDouble                          { height:auto; padding-top:0px;         }
 .RowDouble .TopRow                  { margin:6px 0px 0px; }
-.RowDouble .ZmMsgListExpand         { margin-left:-2px; margin-right:2px; }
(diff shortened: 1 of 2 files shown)


WebRoot/skins/_base/base4/skin.css
```diff
@@ -715,6 +715,15 @@ SELECT                      {   @NativeSelect@  }
     @Text-unread@
 }
 
+.Row .ZmConvListFrom,
```
silentsakky: can we apply this style directly to ZmConvListFrom? it looks like in every case we are applying this

kalpeshhpatel: If you see few of the lines above (https://github.com/Zimbra/zm-web-client/blob/56d075ca1abe1b7a1db915538303a121a8bfceaf/WebRoot/skins/_base/base4/skin.css#L689), you can see that selector is more specific and not just `ZmConvListFrom` and so we need to follow same way to override it.

Thanks

kalpeshhpatel: Updated: https://github.com/Zimbra/zm-web-client/pull/206/files

PR #216, opened 2017-11-23
Bug 108673: Remove dead code which causes warnings

mss:
I was working on something completely different and got distracted by this warning popping up all the time in the dev console. Nohing bad and urgent, just a bit annoying.

This code was killed as part of [bug 58913](https://bugzilla.zimbra.com/show_bug.cgi?id=58913) in 2011.  Since it is anabiotic for ages now, I just killed it for good.


silentsakky:
Merged in https://github.com/Zimbra/zm-web-client/commit/ad2733304305976dc9d4e52e9ef09660826c922f

PR #220, opened 2017-11-29
'Home' was not translated all the way.

Flums:


silentsakky:
Merged in https://github.com/Zimbra/zm-web-client/commit/17eb74dde3bfe421dbdb44d2816499a112db7915

PR #231, opened 2018-02-16
snickers: NICPS-30: display received time in email list

k-kato:
Problem : display received/sent time of all messages in email list. Before the fix, only (year/)month/day is shown if the message was received yesterday or before.
Fix : 
- If zimbraPrefDisplayTimeInMailList is TRUE, web client displays not only date but also time of a message received yesterday or before in email list.
- Added a setting in Preferences -> Mail tab
- Adjust the width of Received field

Testing done : testing done by developer. Tested with the following configurations:
- All supported languages
- Reading pane: right/bottom
- View: Message/Conversation
- zimbraPrefDisplayTimeInMailList: TRUE/FALSE

Testing to be done by QA : testing needed to be done by QA in PS team
Link:
- Set of PRs : https://github.com/Zimbra/zm-ajax/pull/47, https://github.com/Zimbra/zm-web-client/pull/231, https://github.com/Zimbra/zm-mailbox/pull/525
- RP to develop branch: https://github.com/Zimbra/zm-ajax/pull/46, https://github.com/Zimbra/zm-web-client/pull/227, https://github.com/Zimbra/zm-mailbox/pull/511

TODO: I18N support
- ZmMsg.properties: displayTimeInMailList, timeInMailListChangeRestartHide, timeInMailListChangeRestartShow

### DIFF
--- a/WebRoot/js/zimbraMail/mail/ZmMailApp.js
+++ b/WebRoot/js/zimbraMail/mail/ZmMailApp.js
@@ -286,6 +286,7 @@ function() {
 				ZmSetting.DEDUPE_MSG_TO_SELF,
                 ZmSetting.DEDUPE_MSG_ENABLED,
 				ZmSetting.DISPLAY_EXTERNAL_IMAGES,
+				ZmSetting.DISPLAY_TIME_IN_MAIL_LIST,
 				ZmSetting.GET_MAIL_ACTION,
 				ZmSetting.INITIAL_SEARCH,
 				ZmSetting.MAIL_BLACKLIST,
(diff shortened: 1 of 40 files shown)


DevDef:
This is a long time awaited feature! :-)
Translations are missing, though. I’ve just created the French one as pull request #260.

But, as a Zimbra administrator, I would like to be able to set this property on by default for every user in the default COS. I even wonder why _TRUE_ is not the default value for this property as I’ve never heard someone asking for hiding the time in his mail list, and almost everyone complaining about not being able to see it.

k-kato:
@DevDef 
The reason why the default value is FALSE is for backward compatibility. Time was not shown on previous version of zimbra. If the time appears after the upgrade, a user may be surprised/confused at the change. The cos setting should be changed at the right timing for the system/service.

DevDef:
@k-kato 
> If the time appears after the upgrade, a user may be surprised/confused at the change.

I would rather say surprised/pleased ;-), I heard so much complains about this lack of time information.

> The cos setting should be changed at the right timing for the system/service.

Do you mean that it’s already possible do it?

PR #235, opened 2018-03-02
ZCS-4386

silentsakky:
- upgrade yuicompressor to 2.4.8
- make sure zm-ajax code is copied directly to build directory so main directory remains clear from that clutter
- thirdparty libraries will be fetched using ivy
- move TinyMCE package to webclient

ZCS-1511
- Move ZMsg.properties files to zm-ajax as it is common between admin console and webclient


jiteshsojitra:
Mass changes, review at high level and changes looks good. We can do QA by creating the build.

jiteshsojitra:
Mass changes, review at high level and changes looks good. We can do QA by creating the build.

PR #236, opened 2018-03-07
NICPS-30: added an abililty to show received time in email list

silentsakky:


silentsakky:
@k-kato please verify that there are no issues in cherry picking

PR #237, opened 2018-03-07
ZCS-4390 use google closure compiler instead of yuicompressor

silentsakky:
- yuicompressor is out of date now and development is suspended for the product
- so we are using google closure compiler to minify the javascript files
- fix issues which were found when trying for SIMPLE optimization level but we will be using WHITESPACE_ONLY level for now

### DIFF
--- a/WebRoot/js/externs.js
+++ b/WebRoot/js/externs.js
@@ -1,0 +1,761 @@
+/**
+ * externs.in
+ * This file is mainly used by google closure compiler to identify global variables
+ * defined in the application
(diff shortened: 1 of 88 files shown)


jiteshsojitra:
Looks like full regression would be needed as far as UI is concerned. Saket - did you get a chance to look major affected areas after making these changes locally?

silentsakky:
Yes, anyways we will be running full selenium suite for ZCS-4023, so that testing will cover these changes

PR #239, opened 2018-03-09
ZCS-4388 Upgrade tinymce version in webclient repository

silentsakky:
- Upgrade tinymce version to 4.7.9
- Use minified version of tinymce library
- add target to generate minified version of our custom plugins and use minified files in webclient


jiteshsojitra:
I saw this notice when viewing Zimbra project in Github. Is this fine?

We found potential security vulnerabilities in your dependencies.
Some of the dependencies defined in tinymce-4.2.6/package.json have known security vulnerabilities and should be updated.

jiteshsojitra:
Excellent Saket!

mgarbin:
hi guys consider that tinyMCE version 4.7.9 is not compatible with IE <= 10, as they described in this section : https://www.tinymce.com/docs/get-started/system-requirements/

I've compiled my zm-web-client with version 4.5.x and in production system with 10.000+ works great.
They made some backport for cve and other fixed :
https://github.com/tinymce/tinymce/tree/4.5.x

silentsakky:
Good point !!!
Zimbra's latest version is officially supported on IE 11 and above (including edge).
https://zimbra.github.io/installguides/latest/single.html#System_Requirements

PR #241, opened 2018-03-12
 ZCS-4389 Upgrade patches for tinymce to latest version

silentsakky:
- Remove grunt-cli from package.json and make it a patch for that change
- Regenerate all patches based on new version of tinymce
- Fix Makefile to have sorting in place before applying patches
- Commit built files as well, so webclient will be able to pick it for build creation


jiteshsojitra:
Mass changes, high level it looks good to me. We would be needing lot of testing around TinyMCE to consider composing, inline/external images, tables, styles etc..

PR #242, opened 2018-03-13
ZCS-4389 Upgrade patches for tinymce to latest version

silentsakky:
- upgrade gitignore patch to include more files
- commit built minified files as well


jiteshsojitra:
High level it looks good to me. We would be needing lot of testing around TinyMCE to consider composing, inline/external images, tables, styles etc..

mgarbin:
hi, as i written in other commit : 
Guys consider that tinyMCE version 4.7.9 is not compatible with IE <= 10, as they described in this section : https://www.tinymce.com/docs/get-started/system-requirements/

I've compiled my zm-web-client with version 4.5.x and in production system with 10.000+ works great.
They made some backport for cve and other fixed :
https://github.com/tinymce/tinymce/tree/4.5.x

silentsakky:
Good point !!!
Zimbra's latest version is officially supported on IE 11 and above (including edge).
https://zimbra.github.io/installguides/latest/single.html#System_Requirements

mgarbin:
Ok perfectly, i don't see yet the last sys requirement, thx for the info

PR #248, opened 2018-03-16
ZCS-4023 try to fix build

silentsakky:
- AjxMsg.properties are already part of admin console, so no need to copy it


Prashantsurana:
This is not breaking build. 
As replaces takes care of it.

silentsakky:
Actually I was getting this error
cp: cannot stat 'WebRoot/messages/A*.*': No such file or directory
as we don't have AjxMsg.properties file residing at that location anymore

PR #253, opened 2018-03-22
ZCS-2528

mihaipopahtp:
- adding localeName_id so translations will get picked up in the Preferences>Language menu


silentsakky:
Merged these changes with https://github.com/Zimbra/zm-web-client/pull/222, which is already open so we can create a single build for testing

PR #269, opened 2018-05-23
ZBUG-421: fixing the permission in postinstall script.

Prashantsurana:


scripts/postinst.sh
```diff
@@ -0,0 +1,6 @@
+echo "**** Fixing up the permission ****"
+/opt/zimbra/libexec/zmfixperms
+if [ -d /opt/zimbra/conf/templates/ ]; then
```
silentsakky: can we move this to zmfixperms script?
that way it will be done in a generic way

Prashantsurana: This is not possible in patch. As zmfixperms is part of core package, which is not split enough to ship part of patch.

silentsakky: ok

PR #274, opened 2018-06-01
ZCS-4465 Remove admin common package

silentsakky:
- Remove debug output from pkg builder command
- Removed common admin package
- All admin console related code should be part of admin console repo

### DIFF
--- a/pkg-builder.pl
+++ b/pkg-builder.pl
@@ -84,18 +84,6 @@ my %PKG_GRAPH = (
                      '/opt/zimbra/jetty_base/etc/zimbra.web.xml.in'],
       stage_fun  => sub { &stage_zimbra_mbox_webclient_war(@_); },
    },
-   "zimbra-mbox-admin-common" => {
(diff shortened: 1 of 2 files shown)


pkg-builder.pl
```diff
@@ -186,32 +112,9 @@ ($)
...
+      "--pkg-install=/opt/zimbra/*"
    );
-   if ( $pkg_info->{dir_list} )
-   {
-      foreach my $expr ( @{ $pkg_info->{dir_list} } )
-      {
-         push( @cmd, "--pkg-installs=$expr" );
-      }
-   }
-   if ( $pkg_info->{file_list} )
```
Prashantsurana: Removing this is risky. 
With this packages get the ownership of /opt/zimbra/*  it will make upgrade and removal of the package trouble some. Especially working with rhel. 

We should mention individual files and the directories that can only be used by the package.

silentsakky: Fixed

PR #290, opened 2018-07-21
Bugfix/zcs 5514

naijenn:
-  Tested the following and did not have issues: Disabled forget password feature on domain level but enabled it for individual account. refreshed UI but still in preferences set recovery email related section does not get rendered.
- Fixed: In Preferences-> Account->Password RecoverySettings  preferences are not getting refreshed. Need to do an external refresh.
- Updated messages to be dynamic (ex: recovery code time and code validity)
- Unless new info is provided, Support for client side CSRF was added on previous ticket.


WebRoot/messages/ZmMsg.properties
```diff
@@ -4525,9 +4525,12 @@ passwordRecoveryButtonLogin = Back to Login Page
...
-recoveryEmailMessageResend = We have re-sent an email with the validation code to confirm your recovery email address. Enter the code in the field below and click the associated button to validate this email address. <br><br>Don\u0027t see the email in your Inbox? Check your Junk folder. If you still don't see it, click the \u0027Resend Code\u0027 button below and we will resend the code to the requested email.
+recoveryEmailMessageSent = We have {0} an email with a validation code to confirm your recovery email address. {1}{2}Enter the code in the field below and click the associated button to validate this email address. <br><br>Don\u0027t see the email in your Inbox? Check your Junk folder. If you still don't see it, click the \u0027Resend Code\u0027 button below and we will resend the code to the requested email.
+recoveryEmailMessageAttempts = You have {0} more atempts left.\u0020
```
desouzas: `atempts` typo

naijenn: Changed and adding to next commit.

PR #305, opened 2018-08-29
NICPS-378:Limit upload file for valid extensions

zimsuchitgupta:


WebRoot/js/zimbraMail/share/view/dialog/ZmUploadDialog.js
```diff
@@ -243,6 +243,13 @@ function(files, uploadForm, folder) {
 	}
 };
 
+ZmUploadDialog.prototype.setFileExtensions = function(){
+    var extensionsData = appCtxt.getSettings().getInfoResponse.attrs._attrs.zimbraFeatureBriefcaseAllowedFileExtensions;
+    var array;
```
ajinkya81: can you use a better variable here?

zimsuchitgupta: sure... changing 'array' into 'extensionArray'

PR #308, opened 2018-09-03
ZCS-5690

mihaipopahtp:
- tamil translations (20180903)


reedesl:
Mihai, should these tamil files end with *ta-IN, instead of just *ta ?  seems like there are two variants of Tamil, one specific to India and the other somewhere else (Shrilanka?).

mihaipopahtp:
@reedesl ,
I chose the short *_ta version because Tamil India is is the main language (http://st.unicode.org/cldr-apps/v#/ta/Gregorian/  -  Zimbra also uses  uses _fr for French (instead of fr_FR) and _fr_CA for French Canada, same does _pt/pt_BR... so I thought we can use ta_LK, ta_SG, ta_MY if these flavors are targeted later. 
There is powerful reason to keep them with _ta -  I can have them all changed to ta-IN, no problem. 
I think ta_IN would be a bit more consistent to the other languages since I saw underscore is preffered in language abbreviations. 
Please let me know.

PR #330, opened 2018-10-04
fix for bugzilla-107711

CMazzilli:


silentsakky:
this fix is already sent for verification
https://github.com/Zimbra/zm-web-client/pull/328

PR #356, opened 2018-10-23
ZBUG-724: set empty string if search query is null or undefined

k-kato:
[Problem]
"undefined" is shown in a search field when Calendar or Preferences tab is clicked if zimbraPrefShowSearchString is TRUE.

[Root cause]
Search query was set in the field without null/undefined check.

[Fix]
Set an empty string if a search query is null or undefined.

### DIFF
--- a/WebRoot/js/zimbraMail/core/ZmZimbraMail.js
+++ b/WebRoot/js/zimbraMail/core/ZmZimbraMail.js
@@ -2151,7 +2151,7 @@ function(view) {
 				var stb = appCtxt.getSearchController().getSearchToolbar();
 				if (appCtxt.get(ZmSetting.SHOW_SEARCH_STRING) && stb) {
 					var value = currentSearch ? currentSearch.query : app.currentQuery;
-					value = appName === ZmApp.SEARCH ? "" : value;
+					value = appName === ZmApp.SEARCH ? "" : (value || "");
 					stb.setSearchFieldValue(value + " " || "");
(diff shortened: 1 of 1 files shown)


WebRoot/js/zimbraMail/core/ZmZimbraMail.js
```diff
@@ -2151,7 +2151,7 @@ function(view) {
 				var stb = appCtxt.getSearchController().getSearchToolbar();
 				if (appCtxt.get(ZmSetting.SHOW_SEARCH_STRING) && stb) {
 					var value = currentSearch ? currentSearch.query : app.currentQuery;
-					value = appName === ZmApp.SEARCH ? "" : value;
+					value = appName === ZmApp.SEARCH ? "" : (value || "");
```
tarusharora: @k-kato : This change might lead to a side-effect of showing a blank space instead of the placeholder text while coming to 'Calendar' or 'Preferences' tab. This can be done safely by replacing next line of code with `stb.setSearchFieldValue(value ? value + " " : "");`

k-kato: @tarusharora thank you for your comment. The fix on https://github.com/Zimbra/zm-web-client/pull/392 is better.

k-kato:
It is now addressed on https://github.com/Zimbra/zm-web-client/pull/392. Closed.

PR #362, opened 2018-10-26
ZBUG-673 CSRF: HTML mobile - Advanced Search has vulnerability.

silentsakky:
- add crumb check for advanced search form submit request


WebRoot/m/mosearch
```diff
@@ -19,6 +19,7 @@
     <c:set var="action" value="${zm:cook(empty param.paction ? param.action : param.paction)}" scope="request"/>
     <c:choose>
         <c:when test="${not empty param.actionSearch}">
+            <zm:checkCrumb crumb="${param.crumb}"/>
```
plobbes: I get the feeling the crumb check should happen before the choose (always), but maybe I'm missing something?  In theory, it's possible to navigate/GET directly to either of these directly:
- https://your.zcs.svr/m/mosearch
- https://your.zcs.svr/m/zmview

silentsakky: it's added inside choose to make sure it is only happening when form submit happens for search, most of the GET requests in mobile client doesn't have crumb check, so if needed we can handle in a separate ticket

plobbes: OK.  I think this should suffice for now.  We can revisit later if necessary.

PR #363, opened 2018-10-29
 NICPS-430 Hindi Language Support in tinymce

silentsakky:
- Integration of pramukhime plugin with tinymce
- Fix thirdparty plugin to enable functionality when domain name contains . (DOT)


tiny_mce/custom/plugins/pramukhime/langs/hi_pramukhindic.js
```diff
@@ -0,0 +1,44 @@
+// This file should be used only for keyboard plugin specific translation.
```
jkester1986: will eventually need to to translations here, but this is fine for now

PR #369, opened 2018-11-12
ZBUG-732: fixed to check if an email adress is not empty in Add Filter dialog

k-kato:
[Problem]
Filter can be saved even when an email address field of From/To/Cc/Bcc is empty in Add Filter dialog.

[Root cause]
Empty check did not exist for a rule of TEST_ADDRESS type.

[Fix]
Added empty checks.


k-kato:
Please also check if it was a bug or by design. Can we add empty check?

silentsakky:
Closing PR as ticket got marked as won't fix ZBUG-732, reopen if you feel this needs to be merged to develop

k-kato:
I reopened the PR and Jira ticket.  I strongly want a reason to be written in Jira before it is marked as "won't fix".

PR #371, opened 2018-11-13
Fix TypeError

barrydegraaff:
Fixes Uncaught TypeError: Cannot set property 'disabled' of null at ZmComposeView.enableInputs


barrydegraaff:
@silentsakky I have been getting this type error for a while, but cannot reproduce it reliable. What happens in most cases a user would click New message (nothing happens because of the type error). Then the next time the user clicks to New message it works.

This PR solves the type error, so that the UI works correctly.

PR #374, opened 2018-11-14
NICPS-468 : Provide option/dropdown to switch language on Login Page

kvc2983:
- Provide option/dropdown to switch language on Login Page


jkester1986:
@kvc2983 we should really hold off on merging until all PRs are approved and ticket passes CR

PR #383, opened 2018-11-22
NICPS-469: Added language parameter work, change language as per url …

deepakzimb:
- User can select language from login page
- Whatever language selected by user, it should applied on url as lang=<LANG>
- After login, same language applied in Inbox section
- Once user update Preference -> Language, its lang parameter will update in url and Inbox section language should update
- When user click on Sign Out, selected language should be on login page (After logout)


WebRoot/h/search
```diff
@@ -5,13 +5,15 @@
...
 <%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions" %>
 <%@ taglib prefix="fmt" uri="com.zimbra.i18n" %>
-
 <app:handleError>
     <zm:getMailbox var="mailbox"/>
     <c:choose>
     <c:when test="${not empty mailbox.prefs.locale}">
         <fmt:setLocale value='${mailbox.prefs.locale}' scope='request' />
     </c:when>
+    <c:when test="${not empty param.lang}">
```
jkester1986: looks like this new code only runs during errors? Since it's within `<app:handleError>`. Though I may be misunderstanding what is happening here

deepakzimb: This code is simple updating Locale based on condition.

jkester1986: this code should replace the `<c:when/>` below it, they are both setting the locale, so the second one is actually going to be the one that takes effect

PR #390, opened 2018-12-04
ZBUG-623 : Prevent attachment duplication while saving or sending email.

BhavinBathani:
Inline pdf attachments are being converted to actual attachment as we are
considering inline-pdf as non renderable with html editor mode as well.
Fixing this by considering inline-pdf as non-renderable in case of
Text editor only.


mss:
This probably caused [bug 109103](https://bugzilla.zimbra.com/show_bug.cgi?id=109103) (cf. case 00796203).

PR #391, opened 2018-12-05
NICPS-496 : Fix spacing issue for search bar text

tarusharora:
NICPS-496 : Fix spacing issue for search bar text.
Earlier, "undefined" was shown in the search field when Calendar or Preferences tab is clicked.
It was due to the concatenation of undefined variable and a space character.


zimsuchitgupta (approved):
Changes looks good to me!

PR #399, opened 2018-12-13
NICPS-506: fix some language plugin issues

jkester1986:


WebRoot/js/zimbraMail/share/view/htmlEditor/ZmHtmlEditor.js
```diff
@@ -838,6 +838,22 @@ function(id, autoFocus) {
 		menubar: false,
 		ie7_compat: false,
 		object_resizing : true,
+		pramukhime_options: {
```
jkester1986: added this for the signature tinymce (because there are two tinymce init objects)

PR #412, opened 2019-01-11
NICPS-516 : Language translation in search bar

vishnukantjangid:
NICPS-516 : Language translation in search bar

### DIFF
--- a/WebRoot/js/zimbraMail/share/controller/ZmSearchController.js
+++ b/WebRoot/js/zimbraMail/share/controller/ZmSearchController.js
@@ -810,6 +810,13 @@ function(params) {
 			queryString = this._currentQuery || "";
 		}
 
+		if(queryString && queryString.indexOf(":") != -1) {
+			var queryIndex = queryString.indexOf(" ");
+			var updatedQueryString = langSearchArray[queryString.substring(0, queryIndex)]+" "+ queryString.substring(queryIndex);
+			if(updatedQueryString) {
+				queryString = updatedQueryString;
+			}
+		}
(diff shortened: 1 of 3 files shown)


WebRoot/js/zimbraMail/share/controller/ZmSearchController.js
```diff
@@ -810,6 +810,13 @@ function(params) {
 			queryString = this._currentQuery || "";
 		}
 
+		if(queryString && queryString.indexOf(":") != -1) {
```
jkester1986: I know why you're doing your comparison this way, but could you please add a response to my comment on the ticket, that just says that != is intentional and why?

This way we have a record of it if anyone ever needs to refer to it.

Also, it's helpful to respond to a CR request ahead of time with a reason you aren't doing something, so we don't have to ask if something got overlooked when making CR changes - this is especially useful if you aren't online at the same time as your code reviewer

vishnukantjangid: If we are comparing same data types then, ==/!= and ===/!== work as same. But if we are comparing different data types then we prefer ===/!== instead of ==/!=.

If comparing same data types

=! check data type and compare equality
==! check data type and compare equality

If comparing different data types

=! check data type and conversion of data type and compare equality
==! check data type and compare equality

Here we are comparing same data types integer so I use != for comparison. Replacing ==/!= with ===/!== increases the size of the js file, it will take more time to load.

Thank you @jkester1986  suggested me. I will also use ===/!== instead of ==/!= next time.

PR #441, opened 2019-02-13
ZCS-6680 : When a single calendar is selected new events will be created in the calendar selected instead of the default calendar

auminpatel:
When selected a single calendar from the list of calendar from the list of calendar and creating a new event the add dialog selects the calendar that is selected .
If multiple calendar are selected and a new event is been created the add dialog chooses the default calendar.
If none of the calendar is been selected and a new event is been created the add dialog chooses the default calendar.


WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
```diff
@@ -1624,7 +1623,7 @@ function(startDate, duration, folderId, mailItem) {
 				this.defaultAccount = appCtxt.isFamilyMbox ? this.mainAccount : this.visibleAccounts[1];
 				newAppt.setFolderId(calId);
```
silentsakky: maybe this code is unused, but calId is undefined here

PR #443, opened 2019-02-18
ZBUG-747 cannot login to ajax client on Edge 44

silentsakky:
* make sure to add elements in DOM before calling getComputedStyle, as Edge 44 become stricter and doesn't return any values from getComputedStyle when it is not part of DOM hierarchy

### DIFF
--- a/WebRoot/js/zimbraMail/calendar/view/ZmCalColView.js
+++ b/WebRoot/js/zimbraMail/calendar/view/ZmCalColView.js
@@ -2056,8 +2056,8 @@ function(gridEl, allDayEl, i, data, numCols) {
 			allDayEl.appendChild(divEl);
 		} else {
 			// position/size once right here!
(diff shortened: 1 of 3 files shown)


k-kato:
I searched code which calls `Dwt.setBounds` before the specified element is added to DOM, and found suspicious codes:
https://github.com/Zimbra/zm-web-client/blob/8.8.11/WebRoot/js/zimbraMail/calendar/view/ZmCalDayView.js#L777
https://github.com/Zimbra/zm-web-client/blob/8.8.11/WebRoot/js/zimbraMail/calendar/view/ZmCalColView.js#L2059

The former causes DwtException at Day View in Calendar.
1. click Calendar tab
2. click Day button
3. DwtException is shown in Developer Tools of the browser (Edge 44)

I could not find (=reproduce) any path to call the latter code, but we should check if it is actually never called.

k-kato:
https://github.com/Zimbra/zm-ajax/pull/75

silentsakky:
Thanks, fixed it
@miteshsavani found same issue and filed ZCS-6760, but we will fix it here only

PR #458, opened 2019-03-19
ZBUG-546 : As the urls were getting truncated have in the source code in HTML format 

auminpatel:
Have made covert_urls to false has it was truncating the url when added in source code in html format  and even checked ZBUG-1278 is also working as expected.

### DIFF
--- a/WebRoot/js/zimbraMail/share/view/htmlEditor/ZmHtmlEditor.js
+++ b/WebRoot/js/zimbraMail/share/view/htmlEditor/ZmHtmlEditor.js
@@ -820,7 +820,7 @@ function(id, autoFocus) {
 		object_resizing : true,
         font_formats : fonts.join(";"),
         fontsize_formats : AjxMsg.fontSizes || '',
-		convert_urls : true,
+		convert_urls : false,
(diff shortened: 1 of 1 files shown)


WebRoot/js/zimbraMail/share/view/htmlEditor/ZmHtmlEditor.js
```diff
@@ -820,7 +820,7 @@ function(id, autoFocus) {
 		object_resizing : true,
         font_formats : fonts.join(";"),
         fontsize_formats : AjxMsg.fontSizes || '',
-		convert_urls : true,
+		convert_urls : false,
```
BhavinBathani: Did you check and confirm that this would cause any regression ?
This particular change was actually introduced here in https://jira.corp.synacor.com/browse/ZCS-1278

auminpatel: Have checked the working of ZBUG-1278  and it is working as expected from the ticket ZCS-1278
not causing any regression

PR #472, opened 2019-04-15
NICPS-569: Fixed zimbra text display issue in header tab

deepakzimb:
Fixed zimbra text display issue in the browser tab

When language is different than en, hi, ta, then Zimbra labels show


vishnukantjangid:
Looks good.

PR #477, opened 2019-04-23
Fixes dialog overflow outside browser window

barrydegraaff:
A little know feature in Zimbra is that one can set filters that automate email replies. But this can only be done in the CLI.

/opt/zimbra/bin/zmmailbox -z -m myaccount@example.com afrl "Filter_name" active any address "to,cc,from" all is "otheraccount@example.com" reply "Auto reply message" stop

The filter is now shown read-only to the user, however if a longer reply message is set, the filter dialog goes outside the browser window like so:

without-patch

This PR patches it to be like so:
with-patch

Also, without patch the UI can still be used by dragging around the filter dialog, which makes the dialog  resize correctly eventually. But most users would not expect that.


barrydegraaff:
@silentsakky can you set this up for review, thanks!

barrydegraaff:
@Prashantsurana this one seems to be stuck

Prashantsurana:
@silentsakky @dajay0  can you please make sure this get reviewed on the highest priority?

PR #481, opened 2019-04-26
Enforce functional sanitation of "skin" parameter

basilgello:
Many attempts to exploit zimbra "skin" directory traversal bug (see https://www.exploit-db.com/exploits/30085 or its rewritten version https://github.com/nulsec/zimbra-0day/blob/master/zimbra.txt) produce the tremendous amount of exceptions in zmmailboxd.out.

This commit chain closes https://bugzilla.zimbra.com/show_bug.cgi?id=109156 by enforcing functional
sanitation of "skin" parameter. As "skin" parameter is one alphanumeric word. the proposed patches revert to Zimbra's "defaultSkin" option if supplied value is not the case.

### DIFF
--- a/WebRoot/WEB-INF/tags/infra/skinAndRedirect.tag
+++ b/WebRoot/WEB-INF/tags/infra/skinAndRedirect.tag
@@ -15,6 +15,7 @@
  * ***** END LICENSE BLOCK *****
 --%>
 <%@ tag import="java.io.*" %>
+<%@ tag import="java.util.regex.Pattern" %>
 <%@ tag import="javax.servlet.*" %>
 <%@ tag import="javax.servlet.jsp.*" %>
 <%@ attribute name="mailbox" rtexprvalue="true" required="false" type="com.zimbra.cs.taglib.bean.ZMailboxBean" %>
@@ -27,6 +28,10 @@
 	<%
 		PageContext pageContext = (PageContext)jspContext;
 		String skin = (String)pageContext.findAttribute("skin");
+
+		if (!Pattern.matches("^[0-9A-Za-z]+$", skin)) {
+		    skin = application.getInitParameter("zimbraDefaultSkin");
+		}
 
 		if (uri == null) {
 			uri = request.getRequestURI();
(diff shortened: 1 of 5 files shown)


basilgello:
@silentsakky please have a look at the bugzilla discussion and this PR. Thanks in advance!

WebRoot/WEB-INF/tags/infra/skinAndRedirect.tag
```diff
@@ -28,6 +29,10 @@
 		PageContext pageContext = (PageContext)jspContext;
 		String skin = (String)pageContext.findAttribute("skin");
 
+		if (!Pattern.matches("^[0-9A-Za-z]*$", skin)) {
```
mss: Shouldn't this be a `+` instead of a `*`?

basilgello: Yes, you are right. * expands to "zero or more times", while we need "one or more times", exclusing empty string. I will fix the PR now.

basilgello: Check this one please @mss

mss:
LGTM.

Small nit: I'd put the check
```
skin != null && Pattern.matches("^[0-9A-Za-z]+$", skin)
```
into a helper function but I don't know enough about how taglibs work to suggest where that should be put.

basilgello:
@mss The function can be simply wrapped at line 30or in tld definitions, but I don't see the point introducing the helper function for a one-shot comparison in a single tag file.

silentsakky:
@barrydegraaff Because of other urgent priorities this is getting delayed, rest assured we will verify and merge the fix

Prashantsurana:
@silentsakky @dajay0 @mogsdad  Please review the change ASAP and if all okay merge to 8.8.15

WebRoot/WEB-INF/tags/infra/skinAndRedirect.tag
```diff
@@ -28,6 +29,10 @@
 		PageContext pageContext = (PageContext)jspContext;
 		String skin = (String)pageContext.findAttribute("skin");
 
+		if (!Pattern.matches("^[0-9A-Za-z]+$", skin)) {
```
silentsakky: any reason to not use already available zm:cook method https://github.com/Zimbra/zm-web-client/blob/09e633765546281efdb3d6ee84f2fe82771064c0/WebRoot/public/launchNewWindow.jsp#L142 ?
and having custom implementation

basilgello: @silentsakky Thanks for reference to cook method! However, the goal of this commit is not only to eradicate directory traversal bug, but also not to trigger additional errors DoS'ing the logs. If a skin method is escaped only, it becomes "%2F..%2F..%2Fetc%2Flocalconfig.xml" throwing the exception of BundleNotFound. My implementation enforces the business logic rule that skin name is one single word, thus narrowing the manipulation space to a non-existent skin name. But non-existent skin name is handled already, producing no additional log entries. So custom implementation here provides closer approximation to the expected logic.

silentsakky: well I want logs to be added when someone passes some weird data and tries to do xss attack, that would be helpful for admins instead of silently ignoring it.

And also we are talking about a very rare scenario where someone would try to pass some bad data in skin parameter that usually happens when someone wants to do XSS attack, not in a day to day affair

So I would prefer to have constancy in code and have errors in log so admins would be aware about the attack

basilgello: Unpatched instance log is attached to initial bug: https://bugzilla.zimbra.com/attachment.cgi?id=68235. This used to happen seberal times a day  as bots scanned the Zimbra instances from different IPs.

Patched instance produces the following log entry in mailbox.log (not zmmailboxd.out!):

`2019-05-24 05:48:40,307 WARN  [qtp1647809929-67268:https:https://xx.xx.xx.xx/res/I18nMsg,AjxMsg,ZMsg,ZmMsg,AjxKeys,ZmKeys,ZdMsg,Ajx%20TemplateMsg.j
s.zgz?v=091214175450&skin=../../../../../../../../../opt/zimbra/conf/localconfig.xml%00] [] misc - Rejecting request containing null character in que
ry string`

so admins still have the alert for this type of attack.

And this is not an XSS attack, it is sort of local inclusion attack (just like XXE).

basilgello: @silentsakky so what's the final decision on this PR?

silentsakky: ok

basilgello: Thanks! :)

PR #484, opened 2019-05-22
Avoid TypeError in ZmComposeView.prototype.enableInputs

barrydegraaff:
Using typeof does not avoid the error. @silentsakky perhaps you can merge it.

### DIFF
--- a/WebRoot/js/zimbraMail/mail/view/ZmComposeView.js
+++ b/WebRoot/js/zimbraMail/mail/view/ZmComposeView.js
@@ -1439,9 +1439,9 @@ ZmComposeView.prototype.enableInputs =
 function(bEnable) {
     DBG.println('draft', 'ZmComposeView.enableInputs for ' + this._view + ': ' + bEnable);
     this._recipients.enableInputs(bEnable);
-    if (typeof(this._subjectField) !== 'undefined') {
+    try {
        this._subjectField.disabled = this._bodyField.disabled = !bEnable;
-    }
+    } catch (err){}
 };
 
 /**


WebRoot/js/zimbraMail/mail/view/ZmComposeView.js
```diff
@@ -1439,9 +1439,9 @@ ZmComposeView.prototype.enableInputs =
 function(bEnable) {
     DBG.println('draft', 'ZmComposeView.enableInputs for ' + this._view + ': ' + bEnable);
     this._recipients.enableInputs(bEnable);
-    if (typeof(this._subjectField) !== 'undefined') {
```
silentsakky: not exactly sure, we should be doing try/catch
can you check if below fix helps?
`if (typeof(this._subjectField) !== 'undefined' && typeof(this._bodyField) !== 'undefined') {`

barrydegraaff:
I am testing your suggestion, as the bug does not happen every day 
it takes me a week or 2 before I can tell if it works. 

The try/catch did resolve the error. Then other errors show up, but those seem 
to not break anything and are only logged via AjxException. 

Kind regards, 

Barry de Graaff 
Zeta Alliance 
Co-founder & Developer 
zetalliance.org | github.com/Zimbra-Community 

Signal: +31 617 220 227 
Fingerprint: 97f4694a1d9aedad012533db725ddd156d36a2d0

barrydegraaff:
Unfortunately the problem I am having, returned when using 
if (typeof(this._subjectField) !== 'undefined' && typeof(this._bodyField) !== 'undefined') 
instead of try-catch. 

The problem is that when I click new mail, sometimes I need to click it twice to make 
it work because of this issue. 

Please consider merging the PR. 

Kind regards, 

Barry de Graaff 
Zeta Alliance 
Co-founder & Developer 
zetalliance.org | github.com/Zimbra-Community 

Signal: +31 617 220 227 
Fingerprint: 97f4694a1d9aedad012533db725ddd156d36a2d0

barrydegraaff:
@Prashantsurana this one seems to be stuck. Maybe a try...catch is not the best resolution, but I cannot seem to make it work with typeOf and if.

PR #485, opened 2019-05-22
ZBUG-1005 fix

barrydegraaff:
Incoming appointment requests via mailview are not applied to the calendar that is listed as default instead the appointments are added to the primary Calendar. Since 8.8.11. See also: https://github.com/Zimbra/zm-web-client/commit/c170de15380e5bf820cfe6341d32e625664cd8c0#diff-8cd4bb295cdcc2bf53e92ae8ae98827dR1372

method: ZmMailMsg.prototype._handleResponseSendInviteReply

To understand what is happening, if Zimbra gets an Calendar invite in the mailbox, and the user clicks accept,
a SendInviteReplyRequest is fired to Zimbra, this takes care of the reply to the sender and a bunch of other things.

Then based on an if statement in _handleResponseSendInviteReply, the appointment is moved to the correct
calendar using an ItemActionRequest. So the Invite always goes to /Calendar first (which is poor behavior, but anyway)

I found that this if statement fails:

 // allow or disallow move logic:
 var allowMove;
 if ((this.acceptFolderId != appCtxt.get(ZmSetting.CAL_DEFAULT_ID)) ||
 (appCtxt.multiAccounts &&
 !this.getAccount().isMain &&
 this.acceptFolderId == appCtxt.get(ZmSetting.CAL_DEFAULT_ID)))
 {
 allowMove = true;
 }
 if (this.acceptFolderId && allowMove && resp.apptId != null) {
 this.moveApptItem(resp.apptId, this.acceptFolderId);
 }

My guess is that this.acceptFolderId should not have been the same as appCtxt.get(ZmSetting.CAL_DEFAULT_ID
but for some reason it sometimes is and then the bug happens. Digging through the code, I cannot see where this.acceptFolderId is initialized, it is probably an anonymous argument, that is either missing or in the wrong place. All the other stuff in the allowMove logic seems to be legacy code from family mailbox era. But I may be wrong there.

This commit removes the whole allowMove logic. And that seems to work well in our testing.

### DIFF
--- a/WebRoot/js/zimbraMail/mail/model/ZmMailMsg.js
+++ b/WebRoot/js/zimbraMail/mail/model/ZmMailMsg.js
@@ -1367,17 +1367,8 @@ function(callback, toastMessage, result) {
 		this._origMsg.folderId = ZmFolder.ID_TRASH;
 	}
 
-	// allow or disallow move logic:
-	var allowMove;
-	if ((this.acceptFolderId != appCtxt.get(ZmSetting.CAL_DEFAULT_ID)) ||
-		(appCtxt.multiAccounts &&
-			!this.getAccount().isMain &&
-			this.acceptFolderId == appCtxt.get(ZmSetting.CAL_DEFAULT_ID)))
-	{
-		allowMove = true;
-	}
-
-	if (this.acceptFolderId && allowMove && resp.apptId != null) {
+	if (this.acceptFolderId && resp.apptId != null) {
+	
 		this.moveApptItem(resp.apptId, this.acceptFolderId);
 	}
 


barrydegraaff:
@silentsakky an exciting one!

barrydegraaff:
@Prashantsurana this one seems to be stuck. And Codacy does not seem to auto-check the contents of the PR. But I have no clue as to why.

auminpatel:
This Create's Issue in the following Scenario

Eg : 3 users 
test 
test1 
test2

Steps: 
- > if test1 shares it's Inbox  with test2 user (Manager privileges)and test1 also shares one of the calendar (Manager privileges) with test2 user.
-> If test1 accepts  the invite to and invitation  on behalf of test2 the invite should ideally  should go to the test2 shared calendar , but after your changes it goes to test1 default calendar .

barrydegraaff:
@auminpatel  Here in Gmail it works fine:

test

To test I ran Zimbra with ?dev=1 and copy pasted this
https://raw.githubusercontent.com/Hivos/zm-web-client/f1e60e17eb5c8799007eb5bdc2d121f616b14461/tiny_mce/custom/plugins/zemoticons/plugin.js
into the browser console.

Maybe the build process does something to cause this? Or are you using the latest version of Gmail?

I do not know where your issues comes from, perhaps it is related to the spam taggin?

PR #493, opened 2019-06-24
ZBUG-1037 : Inline image & signature images getting attached on replying and forwarding of an email

auminpatel:


silentsakky:
Fixed moved to ZBUG-1046

PR #495, opened 2019-06-26
ZBUG-1046 : Infinite attachement since 8.8.12P3 in mail window

auminpatel:


toslan:
Should it work if I change /opt/zimbra/jetty_base/webapps/zimbra/js/zimbraMail/mail/model/ZmMailMsg.js as suggested?
So far does not help in my case. Am I missing something?
Release 8.8.12.GA.3794.UBUNTU14.64 UBUNTU14_64 FOSS edition, Patch 8.8.12_P3.

silentsakky:
Production files are compressed, so that will not work
You can wait for next release which will contain this fix

PR #521, opened 2019-08-23
ZCS-7778 : replaced \uu with u\u which was causing translations loadi…

auminpatel:
…ng issues in certain languages


auminpatel:
Following are the languages in legacy : Following are the languages that need to be tested : French - french , French - Canadian , Polish , Portuguese (Brazil), Portuguese (Portugal) , Vietnamese, Spanish, Romanian, Turkish

PR #524, opened 2019-09-12
PREAPPS-3313 : Login page customizations

kvc2983:
Provide option on legacy login screen to choose Zimbra X from available UI options (screenshot).

After successful login, User will be redirected to ZimbraX UI, if user has selected Zimbra X UI from the dropdown.

After logout from Zimbra X, user will be re-directed to legacy login screen.

### DIFF
--- a/WebRoot/public/zimbrax.jsp
+++ b/WebRoot/public/zimbrax.jsp
@@ -1,0 +1,14 @@
+<%@ page buffer="8kb" autoFlush="true" %>
+<%@ page pageEncoding="UTF-8" contentType="text/html; charset=UTF-8" %>
+<%@ page import="java.util.*,javax.naming.*,com.zimbra.client.ZAuthResult" %>
+<%
+	
+    ZAuthResult authResult = (ZAuthResult) request.getAttribute("authResult");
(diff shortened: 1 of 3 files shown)


WebRoot/public/zimbrax.jsp
```diff
@@ -0,0 +1,14 @@
...
+<%@ page pageEncoding="UTF-8" contentType="text/html; charset=UTF-8" %>
+<%@ page import="java.util.*,javax.naming.*,com.zimbra.client.ZAuthResult" %>
+<%
+	
+    ZAuthResult authResult = (ZAuthResult) request.getAttribute("authResult");
+    pageContext.setAttribute("csrfToken", authResult.getCsrfToken());
+%>
+<script TYPE="text/javascript">
+    localStorage.setItem("csrfToken" , "${csrfToken}");
+    var url = window.location.origin + "/zimbrax/index.html"
```
silentsakky: can we add a TODO comment here, that we need to remove index.html part when jetty configs are in place

silentsakky:
@kvc2983 can you squash commits and rebase it?

PR #546, opened 2019-11-22
ZBUG-1100: Contact Group name is not updated in the contact list after renamed or deleted

madhavdhuppe:
Fixed issue i.e Contact Group name is not updated in the contact list after renamed or deleted.

### DIFF
--- a/WebRoot/js/zimbraMail/abook/controller/ZmContactController.js
+++ b/WebRoot/js/zimbraMail/abook/controller/ZmContactController.js
@@ -283,6 +283,8 @@ function() {
 ZmContactController.prototype._setViewContents =
 function() {
 	var cv = this._contactView;
+	// We need to extract current i.e editing contact list and keep in this object while setting edit view, Because contact list of editing contact group can get changed from cache while changing contact lists in edit view.
+	this.editingContactList = this._contact.list;
 	cv.set(this._contact, this._contactDirty);
 	if (this._contactDirty) {
 		delete this._contactDirty;
@@ -360,6 +362,9 @@ ZmContactController.prototype._saveListener = function(ev, bIsPopCallback) {
 	view.enableInputs(false);
 
 	var contact = view.getContact();
+	if(this.editingContactList) {
+		contact.list = this.editingContactList;
+	}
 	if (mods && AjxUtil.arraySize(mods) > 0) {
 
 		// bug fix #22041 - when moving betw. shared/local folders, dont modify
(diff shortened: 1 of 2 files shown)


WebRoot/js/zimbraMail/abook/model/ZmContact.js
```diff
@@ -426,9 +426,12 @@ function(node, args) {
 			contact._parseTagNames(node.tn);
 		}
 		AjxUtil.hashUpdate(contact.attr, node._attrs);	// merge new attrs just in case we don't have them
-		contact.list = args.list || new ZmContactList(null);
-		contact._list = {};
-		contact._list[contact.list.id] = true;
+		var arglist = args.list || new ZmContactList(null);
```
BhavinBathani: Can't we have something like following?

```
if(!contact.list) {
     contact.list = args.list || new ZmContactList(null);
}

contact._list = contact._list || {};
contact._list[contact.list.id] = true;
```

madhavdhuppe: as per now we can't have like this.
looking for any alternative and robust way to fix contact deletion

WebRoot/js/zimbraMail/abook/model/ZmContact.js
```diff
@@ -427,7 +427,7 @@ function(node, args) {
 		}
 		AjxUtil.hashUpdate(contact.attr, node._attrs);	// merge new attrs just in case we don't have them
 		contact.list = args.list || new ZmContactList(null);
-		contact._list = {};
+		contact._list = contact._list || {};
```
madhavdhuppe: We were flushing the contact._list array which contains lists in which contact is present.

WebRoot/js/zimbraMail/abook/controller/ZmContactController.js
```diff
@@ -283,6 +283,7 @@ function() {
 ZmContactController.prototype._setViewContents =
 function() {
 	var cv = this._contactView;
+	this.editingContactList = this._contact.list;
```
madhavdhuppe: We need to extract current i.e editing contact list while setting edit view. Because contact list can get changed from cache  while adding contacts to contact group.

WebRoot/js/zimbraMail/abook/controller/ZmContactController.js
```diff
@@ -360,6 +361,9 @@ ZmContactController.prototype._saveListener = function(ev, bIsPopCallback) {
 	view.enableInputs(false);
 
 	var contact = view.getContact();
+	if(this.editingContactList) {
+		contact.list = this.editingContactList;
```
madhavdhuppe: set extracted list to contact.

PR #558, opened 2020-01-07
PREAPPS-3911 : Mobile users taken to deprecated mobile login on Z8 back end when default set to "Zimbra X"

auminpatel:
Mobile users will be moved to zimbrax if they don't select any option from dropDown(if zimbrax is present) and if zimbrax is supported they will be moved to default login page instead of touch-client in mobile.

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -91,6 +91,12 @@
 %>
 <c:set var="zimbraxSupported" value="<%=zimbraxSupported%>" />
 
+<!-- Adding the below code to support small screens and to have default option to zimbrax if zimbrax is Supported-->
+<c:if test ="${(zimbraxSupported eq true) and (touchSupported eq true)}">
+    <c:set var="touchSupported" value="false" />
+    <c:set var="mobileSupported" value="true" />
+</c:if>
+
 <c:catch var="loginException">
 	<c:choose>
 		<c:when test="${(not empty param.loginNewPassword or not empty param.loginConfirmNewPassword) and (param.loginNewPassword ne param.loginConfirmNewPassword)}">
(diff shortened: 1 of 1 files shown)


WebRoot/public/login.jsp
```diff
@@ -116,7 +116,7 @@
                     <c:redirect url="${logoutRedirectUrl}"/>
                 </c:when>
                 <c:when test="${touchSupported and touchLoginPageExists and (empty param.client or param.client eq 'touch') and
-                    (empty param.virtualacctdomain) and (empty virtualacctdomain)}">
+                    (empty param.virtualacctdomain) and (empty virtualacctdomain) and zimbraxSupported ne true}">
```
silentsakky: `and (zimbraxSupported ne true)`
for consistancy

WebRoot/public/login.jsp
```diff
@@ -232,6 +232,9 @@
                             <c:set var="client" value="mobile"/>
                         </c:if>
                         <c:if test="${empty client or client eq 'preferred'}">
+                            <c:if test="${(mobileSupported or touchSupported) and zimbraxSupported}">
+                                <jsp:forward page="/public/zimbrax.jsp"/>
+                            </c:if>
                             <c:set var="client" value="${requestScope.authResult.prefs.zimbraPrefClientType[0]}"/>
```
silentsakky: shouldn't we just need to update client type to zimbrax in this case instead of doing forwarding?

WebRoot/public/login.jsp
```diff
@@ -394,7 +397,7 @@ if (application.getInitParameter("offlineMode") != null) {
 	</c:redirect>
 </c:if>
 
-<c:if test="${(empty param.client or param.client eq 'touch') and touchSupported and touchLoginPageExists}">
+<c:if test="${(empty param.client or param.client eq 'touch') and touchSupported and touchLoginPageExists and zimbraxSupported ne true}">
```
silentsakky: this gives me an impression that we are ignoring client type passed from somewhere else, can we change client type only instead of doing this?

auminpatel: Done

WebRoot/public/login.jsp
auminpatel: Done

WebRoot/public/login.jsp
auminpatel: removed this code and setting the client has zimbrax

WebRoot/public/login.jsp
auminpatel: Have done the checking condition at top

WebRoot/public/login.jsp
```diff
@@ -91,6 +91,12 @@
 %>
 <c:set var="zimbraxSupported" value="<%=zimbraxSupported%>" />
 
+<!-- Adding the below code to support small screens and to have default option to zimbrax if zimbrax is Supported-->
+<c:if test ="${(zimbraxSupported eq true) and (touchSupported eq true)}">
```
auminpatel: Had missed the dollar sign , so the condition failed

PR #561, opened 2020-01-21
NICPS-721: Stop modify pref request for language option

deepakzimb:
Stop modify pref request for language option


vishnukantjangid:
Code looks good.

PR #564, opened 2020-01-28
PREAPPS-4182:Change login options available to the user

madhavdhuppe:
-Implemented functionality to display only classic and modern options on the login page.
-Removed Standard (HTML), Mobile and Touch options from login page
-Removed Webroot/m folder.
-Covered PREAPPS-4210


WebRoot/public/login.jsp
```diff
@@ -276,51 +270,9 @@
...
-                                        <c:if test="${not fn:contains(ignoredQueryParams, testKey)}">
-                                                <c:param name="${p.key}" value='${value}'/>
-                                            </c:if>
-                                        </c:forEach>
-                                    </c:forEach>
-                                </c:redirect>
-                            </c:when>
-                            <c:when test="${client eq 'zimbrax' and zimbraxSupported}">
+                            <c:when test="${client eq 'modern' and zimbraxSupported}">
                                     <jsp:forward page="/public/zimbrax.jsp"/>
```
BhavinBathani: Can't this be modern as well ?

WebRoot/h/login
```diff
@@ -2,13 +2,6 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 
 <%-- Pass on all parameters for deeplinking --%>
-<c:url value="search" var="url">
```
BhavinBathani: instead of removing this can't we just remove "/h/login" from web.xml

madhavdhuppe: It's not working that way

WebRoot/public/login.jsp
madhavdhuppe: done.

madhavdhuppe:
> remove all touchclient related code from login.jsp

Initially i thought it was getting used in PREAPPS-4183 i.e reskin login page, Removed

WebRoot/h/login
silentsakky: yeah it is very tightly coupled so hard to do now

PR #577, opened 2020-02-27
PREAPPS-4408: Switch To HTML client option appears on Loading page and at user dropdown in Zimbra9X

madhavdhuppe:
-Removed Switch To HTML client option from loading page and user dropdown.


WebRoot/skins/_base/base3/splash.html
```diff
@@ -38,9 +38,6 @@ <h1><div id='ZLoginBannerImage' class='ImgLoginBanner' onclick='showCompanyUrl()
 				<div class="message">
 					<script>document.write(ZmMsg.splashScreenLoading)</script>
 				</div>
-				<div class="switch" id="splashScreenSwitchContainer">
-					<script>if(!window.isOffline) { document.write(ZmMsg.splashScreenSwitch) }</script>
```
BhavinBathani: if ZmMsg.splashScreenSwitch is not being used anywhere else then please remove the same as well.

madhavdhuppe: Removed the "splashScreenSwitch" && "htmlClient" translations.

PR #579, opened 2020-03-09
PREAPPS-4279: Enable Rolling Upgrades for Zimbra 9

yashrajpchavda:
- Removed Cancel button from 2FA flow as it was not present in the legacy and was added during re-skin, and was causing some problems
- Fixed an issue to respect the default client preference of the user who belongs on the servers which are not upgraded yet


yashrajpchavda:
Closing this as https://github.com/Zimbra/zm-web-client/pull/580 contains the changes done as part of this PR.

PR #587, opened 2020-05-28
ZBUG-1460:- Fixed Cross-Site Scripting vulnerability in Zimbra web mail with account name.

madhavdhuppe:


WebRoot/js/zimbraMail/share/model/ZmZimbraAccount.js
```diff
@@ -110,7 +110,7 @@ function(name) {
 	var identity = this.getIdentity();
 	// TODO: If no identity and name is set, should create one!
 	if (!identity) return;
-	identity.name = name;
+	identity.name = AjxStringUtil.htmlEncode(name);
```
silentsakky, line 113: we should do html encoding only in the view layer, data should never be modified

WebRoot/js/zimbraMail/share/model/ZmZimbraAccount.js
```diff
@@ -123,7 +123,7 @@ function() {
 	var identity = this.getIdentity();
 	var name = (!identity)
 		? this.settings.get(ZmSetting.DISPLAY_NAME)
-		: identity.name;
+		: AjxStringUtil.htmlDecode(identity.name);
```
silentsakky, line 126: htmlDecode should not be done by us, it's responsibility of browser

WebRoot/js/zimbraMail/share/model/ZmZimbraAccount.js
madhavdhuppe: but i think the the svg onload is getting  executed while sending requests

WebRoot/js/zimbraMail/share/model/ZmZimbraAccount.js
madhavdhuppe: I have some places that we are using AjxStringUtil.htmlDecode()

WebRoot/js/zimbraMail/share/model/ZmZimbraAccount.js
madhavdhuppe: updated

PR #588, opened 2020-05-29
docker_file Create docker file for use in zimbra cloud

silentsakky:


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM busybox
+
+WORKDIR root
+
+# copy zimbra in webapps
+COPY build/dist/jetty/webapps/zimbra.war /opt/zimbra/jetty_base/webapps/zimbra/
```
jrjena136, line 6: can we extract the war here and place file in appropriate dir.

jrjena136:
I think we should keep the same version in the build.xml as it is used in mailbox dockerfile.
zms-classic-webclient:0.0.2.

silentsakky:
> I think we should keep the same version in the build.xml as it is used in mailbox dockerfile.
> zms-classic-webclient:0.0.2.

Sure will update it when merging changes to develop

PR #591, opened 2020-07-07
ZCS-9067 Adding change password rules.

BhavinBathani:
Adding more strict and early verification rules while user is changing password.
Users are shown the password rules that apply to them.
Immediate feedback is given as a new password is entered, in the form of checkmarks next to the given rules.


WebRoot/h/changepass
```diff
@@ -80,63 +83,279 @@
...
+            const uppers = [],
+                lowers = [],
+                numbers = [],
+                punctuations = [],
+                invalidChars = [],
+                invalidPuncs = [];
+
+            const chars = passwordString.split('');
+
+            chars.forEach(char => {
```
silentsakky, line 264: not sure all browsers will be supporting arrow functions

WebRoot/h/changepass
```diff
@@ -80,63 +83,279 @@
...
+                const charCode = char.charCodeAt(0);
+                let isInvalid = false;
+
+                if ("${zimbraPasswordAllowedChars}") {
```
silentsakky, line 268: is this ${ used by jsp?

silentsakky:
- [ ] Same flow should be used when enforcing change password after login, check these validations there as well
- [ ] also after doing forget password, we ask user to change password, check if it is using same screen

WebRoot/h/changepass
```diff
@@ -80,63 +83,279 @@
...
+            checkImg.style.display = "inline";
+        }
+        function unCheck(checkImg, closeImg) {
+            closeImg.style.display = "inline";
+            checkImg.style.display = "none";
+        }
+        function resetImg(condition, checkImg, closeImg){
+            condition ? check(checkImg, closeImg) : unCheck(checkImg, closeImg);
+        }
+        function showPassword(x, y, z) {
```
silentsakky, line 214: better names in place of x,y,z

WebRoot/h/changepass
```diff
@@ -80,63 +83,279 @@
...
+                x.type = "text";   
+                y.style.display = "none";
+                z.style.display = "block";
+            } else {
+                x.type = "password";
+                y.style.display = "block";
+                z.style.display = "none";
+            }
+        }
+        function showOldPassword() {
```
silentsakky, line 225: where are we doing match of old password with new password?

WebRoot/h/changepass
BhavinBathani: yes it is

WebRoot/h/changepass
BhavinBathani: Fixed.

BhavinBathani:
> * [ ]  Same flow should be used when enforcing change password after login, check these validations there as well
> * [ ]  also after doing forget password, we ask user to change password, check if it is using same screen

Both these scenarios are not trivial to fix in this particular ticket.
Let me create separate tickets with some engineering inputs as we discussed already.

WebRoot/h/changepass
BhavinBathani: Fixed.

WebRoot/h/changepass
BhavinBathani: Fixed.

PR #618, opened 2020-12-17
Croatian upgrade

markohruska:
New cro translations.


markohruska:
Missed username, CLA problem.

PR #625, opened 2021-01-20
ZCS-10127:- Do not display external images in Junk(spam in legacy) even with zimbraPrefDisplayExternalImages=TRUE

madhavdhuppe:


WebRoot/js/zimbraMail/mail/view/ZmMailMsgView.js
```diff
@@ -1012,8 +1012,10 @@ function(params) {
 	// - there are <img> tags OR tags with dfbackground set
 	var isSpam = (this._msg && this._msg.folderId == ZmOrganizer.ID_SPAM);
 	var imagesNotShown = (!this._msg || !this._msg.showImages);
-	this._needToShowInfoBar = (!params.isTextMsg &&
-		(!appCtxt.get(ZmSetting.DISPLAY_EXTERNAL_IMAGES) || isSpam) &&
+	
+	// As Per ZCS-10127 we need to always show the info bar if its an spam
```
silentsakky, line 1016: why do we need to check value of DISPLAY_EXTERNAL_IMAGES, if we are hiding images always in spam folder?

aayushvijayvergiya (approved):

silentsakky (approved):

PR #626, opened 2021-01-27
ZCS-9159_1 : Fix response for block common password.

BhavinBathani:
We are not getting zimbraPasswordBlockCommonEnabled attribute in exception response
from server.
We are now using reason which is returned from server instead.

### DIFF
--- a/WebRoot/h/changepass
+++ b/WebRoot/h/changepass
@@ -599,10 +599,13 @@
                     var isRecentlyUsed = true;
                     var jsonResponse = JSON.parse(xmlhttp.responseText);
                     var errorDetails = jsonResponse.Body.Fault.Detail.Error;
+                    var errorReason = jsonResponse.Body.Fault.Reason;
                     if (errorDetails.Code == "account.PASSWORD_RECENTLY_USED") {
                         removeRule("zimbraPasswordEnforceHistory");
                     } else if (errorDetails.Code == "account.INVALID_PASSWORD") {
-                        if (errorDetails.a[0].n == "zimbraPasswordBlockCommonEnabled") {
+                        if (errorDetails.a && errorDetails.a[0].n == "zimbraPasswordBlockCommonEnabled") {
+                            removeRule("zimbraPasswordBlockCommonEnabled");
+                        } else if (errorReason.Text.indexOf('zimbraPasswordBlockCommonEnabled') !== -1) {
                             removeRule("zimbraPasswordBlockCommonEnabled");
                         }
                     } else if (errorDetails.Code == "account.AUTH_FAILED") {


WebRoot/h/changepass
```diff
@@ -599,10 +599,13 @@
...
                     var jsonResponse = JSON.parse(xmlhttp.responseText);
                     var errorDetails = jsonResponse.Body.Fault.Detail.Error;
+                    var errorReason = jsonResponse.Body.Fault.Reason;
                     if (errorDetails.Code == "account.PASSWORD_RECENTLY_USED") {
                         removeRule("zimbraPasswordEnforceHistory");
                     } else if (errorDetails.Code == "account.INVALID_PASSWORD") {
-                        if (errorDetails.a[0].n == "zimbraPasswordBlockCommonEnabled") {
+                        if (errorDetails.a && errorDetails.a[0].n == "zimbraPasswordBlockCommonEnabled") {
+                            removeRule("zimbraPasswordBlockCommonEnabled");
+                        } else if (errorReason.Text.indexOf('zimbraPasswordBlockCommonEnabled') !== -1) {
```
silentsakky, line 608: why we are doing separate logic for this specific scenario? aren't we getting this information as part of errorDetails.a[0].n ?

BhavinBathani: @chinmay15 : Can you please help me to explain the reason to saket?

chinmay15: `errorDetails.a[0].n` can be used when server sends actual value of the attribute. for example, when minimum length is failed, we send length attribute with it's actual value as argument in exception. like below
```
        int minLength = getInt(acct, cos, entry, Provisioning.A_zimbraPasswordMinLength, 0);
        if (minLength > 0 && password.length() < minLength) {
            throw AccountServiceException.INVALID_PASSWORD("too short",
                    new Argument(Provisioning.A_zimbraPasswordMinLength, minLength, Argument.Type.NUM));
        }

```

in `zimbraPasswordBlockCommonEnabled` case common passwords list may be long, and we can not send all the common passwords in exception attribute, so we send attribute name in plain error message. like below
```
        if (getBoolean(acct, cos, entry, Provisioning.A_zimbraPasswordBlockCommonEnabled, false)
            && commonPasswordFilter.mightContain(password)) {
            throw AccountServiceException
                .INVALID_PASSWORD(Provisioning.A_zimbraPasswordBlockCommonEnabled
                    + " password is known to be too common");
        }
```

BhavinBathani: @chinmay15 raised a PR on backend side https://github.com/Zimbra/zm-mailbox/pull/1116

BhavinBathani: Closing this PR as no changes are required on front-end side.

PR #632, opened 2021-03-31
PREAPPS-5930 : Log into Classic UI in case of IE11.

BhavinBathani:
When IE-11 browser is detected, user should be logged into Classic UI.
On login page, do not show Modern UI in dropdown. It will be Classic and Default.

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -82,7 +82,9 @@
     }
 %>
 <c:set var="modernSupported" value="<%=modernSupported%>" />
-
+<c:if test="${ua.isModernIE}">
+	<c:set var="modernSupported" value="false" />
+</c:if>
 <c:catch var="loginException">
 	<c:choose>
 		<c:when test="${(not empty param.loginNewPassword or not empty param.loginConfirmNewPassword) and (param.loginNewPassword ne param.loginConfirmNewPassword)}">


WebRoot/public/login.jsp
```diff
@@ -699,7 +703,7 @@ if (application.getInitParameter("offlineMode") != null) {
                                             <select id="client" name="client" onchange="clientChange(this.options[this.selectedIndex].value)">
                                                 <option value="preferred" <c:if test="${client eq 'preferred'}">selected</c:if> > <fmt:message key="clientPreferred"/></option>
                                                 <option value="advanced" <c:if test="${client eq 'advanced'}">selected</c:if>> <fmt:message key="clientAdvanced"/></option>
-                                                <c:if test="${modernSupported}">
+                                                <c:if test="${modernSupported and !(ua.isModernIE)}">
```
silentsakky, line 706: we can just update modernSupported condition to be false when user is using IE11, rest everything is handled I guess

BhavinBathani: Fixed.

WebRoot/public/login.jsp
```diff
@@ -82,7 +82,9 @@
     }
 %>
 <c:set var="modernSupported" value="<%=modernSupported%>" />
-
+<c:if test="${ua.isModernIE}">
```
silentsakky, line 85: :100:

silentsakky (approved):

auminpatel (approved):

PR #633, opened 2021-04-13
Fix German "to" field

willirath:
The old "bis" means until as in "nine to five". The new "an" implies a recipient as in "message to somebody".


willirath:
Ping. :)

silentsakky (approved):

PR #640, opened 2021-06-04
ZBUG-2220 : Using escaping while parsing loginErrorCode from url.

BhavinBathani:


WebRoot/public/login.jsp
```diff
@@ -93,7 +93,7 @@
 		</c:when>
 		<c:when test="${param.loginOp eq 'relogin' and not empty param.loginErrorCode}">
 			<zm:logout/>
-			<c:set var="errorCode" value="${param.loginErrorCode}"/>
+			<c:set var="errorCode" value="${fn:escapeXml(param.loginErrorCode)}"/>
```
silentsakky, line 96: check what is difference between fn:escapeXml and zm:cook, I prefer to use zm:cook because escapeXml shows that we are escaping xml but it is not the case

BhavinBathani: Fixed.

silentsakky (approved):

WebRoot/public/login.jsp
```diff
@@ -93,7 +93,7 @@
 		</c:when>
 		<c:when test="${param.loginOp eq 'relogin' and not empty param.loginErrorCode}">
 			<zm:logout/>
-			<c:set var="errorCode" value="${param.loginErrorCode}"/>
+			<c:set var="errorCode" value="${zm:cook(param.loginErrorCode)}"/>
```
silentsakky, line 96: same thing is done in hostedlogin.jsp also, see if we want to fix there as well

BhavinBathani: Fixed.

silentsakky (approved):

PR #651, opened 2021-09-30
ZCS-10918: changed not to add offset twice in custom monthly/yearly repeat

k-kato:
**Problem:**
There is a case that a wrong start day is set automatically when the "second/third/fourth day" is chosen in Custom Repeat dialog.

**Root cause:**
Offset is calculated and added twice.
* at lines 294 and 302 for monthly repeat
* at lines 352 and 365 for yearly repeat

**Change:**
Changed not to add an offset to a date at the second call of `getPossibleStartDate`.


lavegupta (approved):

BhavinBathani (approved):

k-kato:
Fixed a regression in yearly repeat

lavegupta (approved):

PR #657, opened 2021-11-30
ZCS-11065: changed to fetch appointments if they have not been cached

k-kato:
**Problem:**
* Appointments disappear when the appointments are not cached and view is refreshed.

(example of the trigger of refresh)
* change folder color

(cases that the issue happens)
(A) login, go to Calendar immediately and change folder color immediately (before SearchRequest/BatchRequest is called)
(B) enable Calendar app only, login and change folder color immediately 

**Root cause:**
When `ZmCalViewController.prototype._refreshAction` is called with `dontClearCache true`, `this.apptCache.clearCache();` is not called. In this case, appointments are removed from the view and (re)loaded from cache. However, there is a case that appointments shown in the view are not stored in cache. As a result, no item is shown in the view at refresh.

**Fix:**
Add a logic to check if there is a cache (i.e. check if appointments shown in the view are stored in cache) and force `clearCache` if no cache exists.

### DIFF
--- a/WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
+++ b/WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
@@ -4054,6 +4054,12 @@ ZmCalViewController.prototype._refreshAction =
 function(dontClearCache) {
 	DBG.println(AjxDebug.DBG1, "ZmCalViewController: _refreshAction: " + dontClearCache);
 	var forceMaintenance = false;
+
+	if (dontClearCache) {
+		var reminderParams = this._app.getReminderController().getRefreshParams();
+		var mergeKey = this.apptCache._getCachedMergedKey(reminderParams);
+		dontClearCache = Boolean(this.apptCache._cachedMergedApptVectors[mergeKey]);
+	}
 	// reset cache
 	if (!dontClearCache) {
 		this.apptCache.clearCache();


WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
```diff
@@ -4054,6 +4054,12 @@ ZmCalViewController.prototype._refreshAction =
 function(dontClearCache) {
 	DBG.println(AjxDebug.DBG1, "ZmCalViewController: _refreshAction: " + dontClearCache);
 	var forceMaintenance = false;
+
+	if (dontClearCache) {
+		var reminderParams = this._app.getReminderController().getRefreshParams();
+		var mergeKey = this.apptCache._getCachedMergedKey(reminderParams);
+		dontClearCache = Boolean(this.apptCache._cachedMergedApptVectors[mergeKey]);
```
silentsakky, line 4061: not sure if `this.apptCache._cachedMergedApptVectors[mergeKey]` can return {}, if it returns empty object then the condition will fail here

k-kato: First, let's check on it with (B) in Description.
**Prerequisite:**
* enable Calendar only

**Case 1:** appointment(s) exists in Day view
1. login. Calendar is shown automatically.
2. change folder color
3. `this.apptCache._cachedMergedApptVectors[mergeKey]` is `undefined`. Then `dontClearCache` is `false`
4. `this.apptCache.clearCache();` is executed
5. change folder color again
6. `this.apptCache._cachedMergedApptVectors[mergeKey]` is an object of `AjxVector`. It has an array and the array has the appointment(s). `dontClearCache` is `true`
7. `this.apptCache.clearCache();` is not executed

**Case 2:** no appointment exists in Day view
1. login. Calendar is shown automatically.
2. change folder color
3. `this.apptCache._cachedMergedApptVectors[mergeKey]` is `undefined`. Then `dontClearCache` is `false`
4. `this.apptCache.clearCache();` is executed
5. change folder color again
6. `this.apptCache._cachedMergedApptVectors[mergeKey]` is an object of `AjxVector`. It has an empty array. `dontClearCache` is `true`.
7. `this.apptCache.clearCache();` is not executed


`this.apptCache._cachedMergedApptVectors[mergeKey]` is always `undefined` or `AjxVector` object here.
The `_cachedMergedApptVectors` can be set to `{}` by executing `ZmApptCache.clearCache()`.
```
ZmApptCache.prototype.clearCache =
function(folderId) {
	//...
	this._cachedMergedApptVectors = {};
```
Even if `ZmApptCache.clearCache()` is called just before `ZmCalViewController._refreshAction(true)` is called, `this.apptCache._cachedMergedApptVectors[mergeKey]` is `undefined`.

silentsakky (approved):

PR #658, opened 2021-12-01
ZBUG-2553: disabled smooth scrolling in calendar day view on IE11

k-kato:
**Problem:**
Vertical scrollbar does not work by mouse wheel when horizontal scrollbar is moved to the right in calendar Day view on IE11.

**Root cause:**
Smooth scrolling function on IE11 breaks the scroll behavior on web client.

**Fix:**
Catch mouse wheel event and prevent browser's default mouse scroll action (i.e. smooth scrolling) on IE11. Instead, div element is scrolled by the application.


damini-vashishtha (commented):
LGTM

WebRoot/js/zimbraMail/calendar/view/ZmCalDayView.js
```diff
@@ -327,6 +334,13 @@ function(abook) {
 	this._scrollToTime(8);
 };
 
+ZmCalDayView.__onMouseWheel =
+function(elementId, event) {
+	// stop smooth scrolling
+	event.preventDefault();
+	document.getElementById(elementId).scrollTop -= event.wheelDelta / 2;
```
silentsakky, line 341: I am not very sure if browsers optimise this behaviour and not call document.getElementById twice for evaluations
better to get dom element in a variable and then perform operation like below
```
var elm = document.getElementById(elementId);
elm.scrollTop -= event.wheelDelta / 2;
```

WebRoot/js/zimbraMail/calendar/view/ZmCalDayView.js
```diff
@@ -319,6 +319,13 @@ function(abook) {
     // Fix for bug: 66603. Attaching a scroll function.
     document.getElementById(this._tabsContainerDivId).onscroll = func;
 
+	if (AjxEnv.isModernIE) {
```
silentsakky, line 322: not sure if this issue is with IE11 and above only, as this check will only match IE11 and edge versions
and will not handle IE 10 and below versions

silentsakky: although not very important to handle as microsoft itself has stopped supporting older IE versions

WebRoot/js/zimbraMail/calendar/view/ZmCalDayView.js
k-kato: fixed it

silentsakky (approved):

WebRoot/js/zimbraMail/calendar/view/ZmCalDayView.js
k-kato: It matches IE11 only.
For the latest MS Edge based on chromium, it is regarded as Chrome.
```
AjxEnv.isChrome: true
AjxEnv.isIE: false
AjxEnv.isMSEdge: false
AjxEnv.isModernIE: false
```
When using MS Edge based on chromium + IE mode, it is regarded as IE11. It is the same as native IE11.
```
AjxEnv.isChrome: false
AjxEnv.isIE: false
AjxEnv.isMSEdge: false
AjxEnv.isModernIE: true
```
For IE10 or below versions, Zimbra does not support them. In addition, we cannot install IE10 on windows 10 (downgrade install is not allowed.) I didn't write the condition as `AjxEnv.isIE || AjxEnv.isModernIE` because we cannot test it on them and they are out of support of Zimbra.
Theoretically, `AjxEnv.isModernIE` is false on them and then the issue is reproducible on them.

damini-vashishtha (approved):

PR #563, opened 2020-01-23
PREAPPS-4183 : adding login page changes for zimbra9

auminpatel:
Adding zimbra9 login page changes
Adding css changes and adding new template for the two - factor authentication
Changing background image and colour and adding footer message
Updated the client info message for tooltip
Changing password-recovery template to match modern ui and change in css.
Adding color in skin.properties for branding and moving background image and color to a variable

PR #661, opened 2022-01-06
ZBUG-2443: fixed translation of rememberMe for Japanese

k-kato:
Fixed translation of "Stay signed in" checkbox label in login page for Japanese.

Before: このユーザー情報を保存
After: ログイン状態を保持

Note: `rememberMeMobile` is no longer referred from anywhere. It has been removed from login.jsp since ZCS 9.0.0 on https://github.com/Zimbra/zm-web-client/pull/563. ~~However, the PR fixes it as well just to keep consistency with `rememberMe`.~~ The PR removes `rememberMeMobile` key.


silentsakky:
Better to remove rememberMeMobile from translation system if not getting used in the application

k-kato:
Removed `rememberMeMobile` key and updated Description.

silentsakky (approved):

PR #663, opened 2022-01-11
ZBUG-2236: Show delivered From if no good contact fullname

ronstra-synacor:
**Problem:** If a contact exists, its display name is always used for an address, even if the delivered name is more descriptive, e.g. contact says "user1" and the original message had "User One <user1@example.com>".

**Solution:** Use the more descriptive name for To/From displays.

**Fix:** Fixed ZmMailItem.prototype._parseParticipantNode to use the passed From if the contact is only the user part of the email address.

**Testing:** Deploy zm-web-client branch, follow instructions in [ZBUG-2236](https://jira.corp.synacor.com/browse/ZBUG-2236), verify that delivered fullname is used in final step for viewing and replying.

### DIFF
--- a/WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
+++ b/WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
@@ -176,6 +176,11 @@ function(node) {
 			contact = contactList && contactList.getContactByEmail(node.a),
 			fullName = contact && contact.getFullNameForDisplay(false);
 
+		// unless it's just the start of the email address (ZBUG-2236)
+		if (node.p != null && node.a != null && node.a.startsWith(fullName + "@")) {
+			fullName = node.p; // Update contacts with prettier name?
+		}
+
 		var addr = new AjxEmailAddress(node.a, type, fullName || node.p, node.d, node.isGroup, node.isGroup && node.exp);
 		var ac = window.parentAppCtxt || window.appCtxt;
 		ac.setIsExpandableDL(node.a, addr.canExpand);
--- a/WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
+++ b/WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
@@ -177,7 +177,7 @@ function(node) {
 			fullName = contact && contact.getFullNameForDisplay(false);
 
 		// unless it's just the start of the email address (ZBUG-2236)
-		if (node.p != null && node.a != null && node.a.startsWith(fullName + "@")) {
+		if (node.p && node.a && node.a.startsWith(fullName + "@")) {
 			fullName = node.p; // Update contacts with prettier name?
 		}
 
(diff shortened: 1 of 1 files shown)


WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
```diff
@@ -176,6 +176,11 @@ function(node) {
 			contact = contactList && contactList.getContactByEmail(node.a),
 			fullName = contact && contact.getFullNameForDisplay(false);
 
+		// unless it's just the start of the email address (ZBUG-2236)
+		if (node.p != null && node.a != null && node.a.startsWith(fullName + "@")) {
+			fullName = node.p; // Update contacts with prettier name?
```
k-kato, line 181: Remove a comment.
FYI, "p" is personal name according to API reference.
https://files.zimbra.com/docs/soap_api/9.0.0/api-reference/zimbraMail/GetMsg.html

ronstra-synacor: Where are guidelines for this project on what makes a useful comment? I thought this one went along well with the one above it to explain the logic and give a ticket reference for context without searching the logs. Although I think it's useful, I don't have an issue with removing it. I just want to know what kind of comments are expected.

k-kato: We have internal wiki about coding style.
Regarding the PR, the comment at line 179 is useful as a reference. The comment at line 181 is unnecessary because it's clear enough from the code.
Additionally, we don't need to add a reference comment (like line 179) when the change is not complicated because we can find a commit id, a PR and a Jira ticket number using `git blame` command and blame view on github. Otherwise, we will add and see lots of reference comments. It is not good to keep the code readable.
In my opinion, we can remove the comment at 179 for the above reason.

k-kato (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):
LGTM

WebRoot/js/zimbraMail/mail/model/ZmMailItem.js
ronstra-synacor: The first comment is there because the following code contradicts what the _existing_ comment at the beginning of the function above it says.

The second isn't describing that line, but more an XXX or TODO indicating that here's a place where we should maybe consider updating contacts with the prettier name we found. It wasn't describing the simple assignment. Maybe that should be entered as a separate ticket if necessary. For now, I just got rid of it.

Personally, I spend most of my time grepping and reading the source directly outside of GitHub or an IDE, so I appreciate a terse comment preventing me from having to dig through git blame or log then follow the ticket just to see why something was done. That is, I think they make the code more readable and friendly, as long as they're not entirely superfluous.  However, if the policy for the project is minimal comments, that's what I'll do.

rcyarrapothu (approved):

PR #672, opened 2022-02-04
ZCS-11122: Protect against XSS while rendering list view in /h/calendar

jholder85638:
When rendering the basic html client calendar view, protect against XSS using the fn:escapeXML function to render newAppt

### DIFF
--- a/WebRoot/WEB-INF/tags/calendar/monthView.tag
+++ b/WebRoot/WEB-INF/tags/calendar/monthView.tag
@@ -124,7 +124,7 @@
                                 </c:when>
                                 <c:otherwise>
                                     <app:calendarUrl var="monthZoomUrl" view="month" zoom="${true}" timezone="${timezone}" rawdate="${currentDay}"/>
-                                    <td width="11%" onclick='javascript:dayClick(event, "${monthZoomUrl}");' class='ZhCalMonthDay${currentDay.timeInMillis eq date.timeInMillis ? 'Selected':''}'>
+                                    <td width="11%" onclick='javascript:dayClick(event, "${fn:escapeXml(monthZoomUrl)}");' class='ZhCalMonthDay${currentDay.timeInMillis eq date.timeInMillis ? 'Selected':''}'>
                                             <table width="100%" cellspacing="2" >
                                                 <tr>
                                                     <c:set var="T" value="${zm:isSameDate(currentDay, today) ? 'T' : ''}"/>
--- a/WebRoot/WEB-INF/tags/calendar/multiDay.tag
+++ b/WebRoot/WEB-INF/tags/calendar/multiDay.tag
@@ -291,7 +291,7 @@
                             <td valign=top class='ZhCalDayHour' nowrap width="1%" rowspan="4" style='border-left:none;color:blue;'>
                                 <fmt:formatDate var="dateDf" value="${row.date}" pattern="yyyyMMdd'T'HHmmss" timeZone="${timezone}"/>
                                 <app:calendarUrl var="newAppt" timezone="${timezone}" date="${dateDf}" action="edit"/>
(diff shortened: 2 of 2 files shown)


WebRoot/WEB-INF/tags/calendar/monthView.tag
```diff
@@ -124,7 +124,7 @@
                                 </c:when>
                                 <c:otherwise>
                                     <app:calendarUrl var="monthZoomUrl" view="month" zoom="${true}" timezone="${timezone}" rawdate="${currentDay}"/>
-                                    <td width="11%" onclick='javascript:dayClick(event, "${monthZoomUrl}");' class='ZhCalMonthDay${currentDay.timeInMillis eq date.timeInMillis ? 'Selected':''}'>
+                                    <td width="11%" onclick='javascript:dayClick(event, "${fn:escapeXml(monthZoomUrl)}");' class='ZhCalMonthDay${currentDay.timeInMillis eq date.timeInMillis ? 'Selected':''}'>
```
BhavinBathani, line 127: I think we need to check if fn:escapeXml is proper function or zm:cook

silentsakky: I see in many places in tags file we are using fn:escapeXml, so let's go ahead with that as a safer bet

silentsakky (approved):

k-kato (approved):

WebRoot/WEB-INF/tags/calendar/monthView.tag
k-kato: FYI:
* zm:cook  https://github.com/Zimbra/zm-mailbox/blob/develop/common/src/java/com/zimbra/common/util/StringUtil.java#L769
* fn:escapeXml  https://github.com/javaee/jstl-api/blob/master/impl/src/main/java/org/apache/taglibs/standard/tag/common/core/Util.java#L118

silentsakky:
I am assuming this change will be applicable to 8.8.15 only, because in 9.0.0 we have removed /h/calendar endpoint as part of https://github.com/Zimbra/zm-web-client/pull/631 and https://github.com/Zimbra/zm-web-client/pull/628

WebRoot/WEB-INF/tags/calendar/monthView.tag
jholder85638: Generally, escapeHtml is considered much safer than escapeXml.
escapeXml only supports escaping the five basic XML entities (gt, lt, quot, amp, apos) whereas escapeHtml supports escaping all known [HTML 4.0 entities](http://www.w3.org/TR/REC-html40/sgml/entities.html).

In this instance, escapeXML was only chosen based on the previous code I saw.

k-kato: @jholder85638 `zm:cook` or `com.zimbra.common.util.StringUtil#escapeHtml` removes the 5 characters plus CR and LF.

jholder85638:
> I am assuming this change will be applicable to 8.8.15 only, because in 9.0.0 we have removed /h/calendar endpoint as part of #631 and #628

[These are my personal comments, and should not be used an official source of response from Synacor/Zimbra]
This is correct. ZCS 9.0 P15 or later cannot be affected. ZCS 9.0 P14 and earlier included some /h/ endpoints. A review of these by multiple people (myself included) did not find any vulnerabilities. 

Most Zimbra versions prior to 8.8.15 P30 are affected. Users should upgrade.

If users cannot upgrade, they must not apply this fix as it is only designed for 8.8.15. In that case, immediate triage can be achieved by removing /h/calendar altogether. This will break the basic HTML calendar, but does resolve the vulnerability in versions prior to 8.8.15

PR #674, opened 2022-02-08
ZBUG-2084: Prevent Javascript insertion into the error.jsp page on Zimbra9

jholder85638:
On ZCS9, the errorcode field is not properly escaped. This fixes that.

### DIFF
--- a/WebRoot/public/error.jsp
+++ b/WebRoot/public/error.jsp
@@ -64,7 +64,7 @@
 	<c:if test="${empty favIconUrl}">
 		<fmt:message key="favIconUrl" var="favIconUrl"/>
 	</c:if>
-	<link rel="SHORTCUT ICON" href="<c:url value='${favIconUrl}'/>">
+	<link rel="SHORTCUT ICON" href="<c:url value='${fn:escapeXml(favIconUrl)}'/>">
 </head>
 <c:set value="/img" var="iconPath" scope="request"/>
 <body>
--- a/WebRoot/public/error.jsp
+++ b/WebRoot/public/error.jsp
@@ -53,7 +53,7 @@
 <html>
 <head>
 	<meta http-equiv="Content-Type" content="text/html;charset=utf-8">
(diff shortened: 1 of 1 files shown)


silentsakky (approved):

ghen2:
The second and third changes are fmt:message lookup keys, those don't need to be escaped? 
(you're not escaping them in other places either)

jholder85638:
> and third changes are fmt:message lookup keys, those don't need to be

fmt:message does not necessitate escaping. Only in cases where the input can be changed by a data source or user input. Otherwise, most fmt:message tags render mostly zmmsg properties.

In this particular case,
`<title>${fn:escapeXml(errCode)}` 
is escaped because errCode is a variable to the request context: 
`Object errorCode = (Object) request.getParameter("errCode");`

(second) The:
`content="<fmt:message bundle="${zmmsg}" key="zimbraLoginMetaDesc"/>">` is safe because it comes from zmmsg props.

(third) The:
`	<zm:getFavIcon request="${pageContext.request}" var="favIconUrl" />
	<c:if test="${empty favIconUrl}">
		<fmt:message key="favIconUrl" var="favIconUrl"/>
	</c:if>`
The parameter does not need escaping out of the box. favIconUrl is a key in zmmsg props, so it's never (supposed to be) empty. 

However, upon further review, I've decided to escape it anyway. There is a scenario wherein if for some reason the admin edits the webroot /messages/ZmMsg_[lang].properties file and sets favIconUrl to empty, then an attacker can add that to the URL and inject a script. An example would be if the browser sends an unknown character locale, then a zmmsg prop file may not load, thus faviconUrl would test empty. So the attack scenarios are remote, but they should be addressed regardless.

A review of the other keys finds that they are all safe as non will accept GET parameters if zmmsg props cannot be found.

jholder85638:
Please review and let me know of any concerns. Happy to have another set of eyes.

ghen2 (approved):

silentsakky (approved):

PR #684, opened 2022-03-04
ZBUG-2659 Fix XSS security issue on html client

deepakzimb:
Problem: This is a Security issue and users can post HTML/Script content with URL parameters. 

Solution: Added escape-xml method in URL params and sanitized them.

Impact: If the user will pass the HTML/script tag in the URL ads parameter, it will be sanitized, no alert dialog will appear.

### DIFF
--- a/WebRoot/WEB-INF/tags/ads.tag
+++ b/WebRoot/WEB-INF/tags/ads.tag
@@ -22,7 +22,7 @@
 <%@ taglib prefix="zm" uri="com.zimbra.zm" %>
 <%@ taglib prefix="app" uri="com.zimbra.htmlclient" %>
 <c:url var="url" value="/h/ads">
-   <c:param name="f" value="${content}"/>
+   <c:param name="f" value="${fn:escapeXml(content)}"/>
 </c:url>
-<!-- ||${content}|| -->
+<!-- ||${fn:escapeXml(content)}|| -->
 <iframe width="163" height="606" frameborder="0" scrolling="no" src="${url}"></iframe>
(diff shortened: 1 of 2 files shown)


WebRoot/WEB-INF/tags/infra/view.tag
```diff
@@ -519,7 +519,7 @@ if (screen.width<=800) {
 								<td valign="top" colspan="3"></td>
 				<c:if test="${adsOn}" >
 								<td valign="top" style="border-top:1px solid #98ADBE; width:180px;">
-									<app:ads content="${ads}"/>
+									<app:ads content="${fn:escapeXml(ads)}"/>
```
silentsakky, line 522: we usually prefer doing this sanitisation at leaf node, as there could be possibility that `<app:ads />` tag can be used from some other place also and developer missed to sanitise value before passing in the tag
so is it possible that we do this sanitisation inside the ads tag?

deepakzimb: @silentsakky after all findings I found that we have to use fn:escapeXml here and one more file that is ads.tag and I made changes. In infra/view.tag we have to use because it renders HTML in the ads sidebar and that HTML load alert dialog .

shrutig0510 (approved):

WebRoot/WEB-INF/tags/infra/view.tag
```diff
@@ -414,7 +414,7 @@ if (screen.width<=800) {
 	
 	<c:if test="${adsOn}" >
 					<td valign="top" style="border-top:1px solid #98adbe; width:180px;">
-						<app:ads content="${ads}"/>
+						<app:ads content="${fn:escapeXml(ads)}"/>
```
silentsakky, line 417: as discussed, this will do double escaping

silentsakky (approved):

WebRoot/WEB-INF/tags/infra/view.tag
deepakzimb: This is fixed now. Removed from view.tag and added escapexml in leaf file that is ads.tag

dasiyogesh (approved):

preshitaagrawal (approved):

PR #689, opened 2022-04-05
ZCS-11198 : Fixing XSS attack issue with calendarUrl tag.

BhavinBathani:


WebRoot/WEB-INF/tags/imgurl.tag
```diff
@@ -19,4 +19,5 @@
 %><%@ taglib prefix="zm" uri="com.zimbra.zm"
 %><%@ taglib prefix="fmt" uri="com.zimbra.i18n"
 %><%@ taglib prefix="app" uri="com.zimbra.htmlclient"
-%><app:imginfo var="info" value="${value}" /><c:url value="${zm:getImagePath(pageContext, info.src)}" />
\ No newline at end of file
+%><%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions"
+%><app:imginfo var="info" value="${fn:escapeXml(value)}" /><c:url value="${zm:getImagePath(pageContext, info.src)}" />
```
deepakzimb, line 23: Can we sanitize it on imginfo tag file?

BhavinBathani: I think it's not necessary because value attribute in imginfo tag is being used mostly to manage cache only. And is mostly embedded JAVA code.

its seems an invalid URL XSS vulnerability case.
What you say ?

deepakzimb: @silentsakky What do you suggest? Do we have to put it on the leaf node? Because imginfo might be use by other places also.

silentsakky: app:imginfo is used in two places app:img and app:imgurl, so yes better to sanitize at leaf node
although I am not able to understand bhavin's statement, are we suggesting that this is non vulnerable endpoint and we should remove this fix entirely?

deepakzimb: I think we are good after fix on file WebRoot/WEB-INF/tags/infra/appTop.tag and changes are
<form method="get" onsubmit="return searchClick(this);" action="${fn:escapeXml(zm:replaceAll(searchUrl,'javascript:',''))}">

As per the next fix, imginfo one, we can discuss or take this in different ticket. What do you suggest?

silentsakky (approved):

deepakzimb (approved):

PR #691, opened 2022-04-12
ZBUG-2322 : Fix sorting in Tasks.

BhavinBathani:
Problem:
- Sort order selected by user for Task folders is not getting saved/remembered.
- Whatever sort order used by user gets overwritten with default sort order.
- Sort order is being saved in zimbraPrefSortOrder for Task view only.
- Sort order should be saved in zimbraPrefSortOrder for every Task folder.

Solution:
- Sort order is now saved in zimbraPrefSortOrder for every Task folder.
- Different sort order can be saved for different Task folders.


silentsakky (approved):
Indentation looks off in some places, let's fix it

BhavinBathani:
> Indentation looks off in some places, let's fix it

It was wrong earlier.
Anyway, i fixed it again.

deepakzimb (approved):

PR #695, opened 2022-04-25
ZBUG-896: Always add username to logout redirect URL

ronstra-synacor:
**Problem**
During login/logout in classic web UI, the code does not have access to the specific domain of the user.

**Fix**
Send username as parameter of the logout redirect URL, and derive domain from that.

**See Also**
https://github.com/Zimbra/zm-taglib/pull/44

### DIFF
--- a/WebRoot/js/zimbraMail/core/ZmZimbraMail.js
+++ b/WebRoot/js/zimbraMail/core/ZmZimbraMail.js
@@ -2771,9 +2771,9 @@ function(ev, relogin) {
                         loginOp: relogin ? 'relogin' : 'logout'
                     }
                 };
-	if (relogin) {
-		urlParams.qsArgs.username = appCtxt.getLoggedInUsername();
-	}
+	// Formerly, only set below when "relogin" is true.
+	urlParams.qsArgs.username = appCtxt.getLoggedInUsername();
+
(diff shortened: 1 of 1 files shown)


rcyarrapothu (approved):

k-kato (commented):
username should not always be added to url parameter. If it is for handling zimbraWebClientLogoutURL _of a domain_, username should be added only when zimbraWebClientLogoutURL is specified on a domain.

ronstra-synacor:
> It is for security reason. Username must not be passed to login page when the user logs out regardless of implicitly/explicitly. The `relogin` handling is special case for `ZmCsfeException.SVC_AUTH_EXPIRED, ZmCsfeException.SVC_AUTH_REQUIRED or ZmCsfeException.NO_AUTH_TOKEN`.

I don't see a security problem except in an extremely unlikely and not very profitable information-leaking sense, but I'll go with policy.

What about extracting the domain from the username and sending `customerDomain` instead?

k-kato:
customerDomain or username doesn't need to be passed.
zimbraWebClientLoginURL / zimbraWebClientLogoutURL in a domain setting can be applied by using zimbraVirtualHostname.

Regarding domainInfo cache, let's discuss it in https://github.com/Zimbra/zm-taglib/pull/44

PR #700, opened 2022-05-18
ZBUG-2786: Selection of <top> or <bottom> menu items in Insert/edit link dialog of Compose & Event forms breaks the entered URL

luisenriquez01:
- Remove <top> and <bottom> menu items by setting anchor values to false.

Official TinyMCE documentation (v4): https://www.tiny.cloud/docs-4x/configure/url-handling/


WebRoot/js/zimbraMail/share/view/htmlEditor/ZmHtmlEditor.js
```diff
@@ -835,6 +835,8 @@ function(id, autoFocus) {
         font_formats : fonts.join(";"),
         fontsize_formats : AjxMsg.fontSizes || '',
 		convert_urls : false,
+		anchor_top: false,
```
silentsakky, line 838: do we need to fix this in modern ui as well?

luisenriquez01: Yes, it's also reproducible on modern.

silentsakky: ok then let's file similar ticket for modern ui and fix there as well

silentsakky (approved):

PR #703, opened 2022-06-01
ZBUG-1539: blank search : 'include shared items' shows inid:<number> in search query

rababkazi:
On performing a blank search using the topmost search bar with 'Include shared items' checked, search results toolbar was being populated with a query of shared items' ids. Changes have been made in share/view/ZmSearchResultsToolbar.js file to setSearch method to perform a check for the user entered search query. The search results toolbar will be populated with the user entered search query only if it exists.

### DIFF
--- a/WebRoot/js/zimbraMail/share/view/ZmSearchResultsToolBar.js
+++ b/WebRoot/js/zimbraMail/share/view/ZmSearchResultsToolBar.js
@@ -133,7 +133,7 @@ function(search) {
 	this._settingSearch = true;
 	this._searchField.clear(true);
 	var tokens = search.getTokens();
-	if (tokens && tokens.length) {
+	if (search.query && tokens && tokens.length) {
 		for (var i = 0, len = tokens.length; i < len; i++) {
 			var token = tokens[i], prevToken = tokens[i - 1], nextToken = tokens[i + 1];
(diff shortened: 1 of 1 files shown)


k-kato (approved):
LGTM

It is better to write a root cause or the reason why the approach solves the issue in the description of the PR. 
In this case, for example, `queryHint` is set in `ZmSearchController.prototype._doSearch`
```
ZmSearchController.prototype._doSearch =
function(params, noRender, callback, errorCallback) {
	else if (params.inclSharedItems || this._inclSharedItems) {
		// a query hint is part of the query that the user does not see
		params.queryHint = ZmSearchController.generateQueryForShares(type);
	}
```
and it is parsed at ZmSearch if a search toolbar is an empty
```
ZmSearch = function(params) {
			var pq = this.parsedQuery = new ZmParsedQuery(this.query || this.queryHint);
```
and then it is set in `ZmParsedQuery._tokens`
```
ZmParsedQuery = function(query) {
	this._tokens = this._parse(AjxStringUtil.trim(query, true));
```
As a result, not an empty search query but a query hint is returned as tokens.
```
ZmSearchResultsToolBar.prototype.setSearch =
function(search) {
	var tokens = search.getTokens();
```

silentsakky (approved):

PR #707, opened 2022-07-06
ZCS-11202: Fix title sanitization in html

deepakzimb:
Problem: "title" was not properly sanitized.

Solution: Sanitized "title" properly for HTML mode.

### DIFF
--- a/WebRoot/WEB-INF/tags/img.tag
+++ b/WebRoot/WEB-INF/tags/img.tag
@@ -40,7 +40,7 @@
         <c:otherwise> src="${info.src}" </c:otherwise>
     </c:choose>
     <c:choose>
-        <c:when test="${not empty title}"> title='${title}' </c:when>
+        <c:when test="${not empty title}"> title='${fn:escapeXml(title)}' </c:when>
         <c:when test="${not empty alt}"> title='${alt}' </c:when>
     </c:choose>
     <c:if test="${not empty alt}"> alt="${fn:escapeXml(alt)}" </c:if>


WebRoot/WEB-INF/tags/img.tag
```diff
@@ -28,7 +28,7 @@
 <%@ taglib prefix="fmt" uri="com.zimbra.i18n" %>
 <%@ taglib prefix="zm" uri="com.zimbra.zm" %>
 <c:if test="${not empty altkey}"><fmt:message key="${altkey}" var="alt"/></c:if>
-<c:if test="${not empty title and not rawtitle}"><fmt:message key="${title}" var="title"/></c:if> 
+<c:if test="${not empty title and not rawtitle}"><fmt:message key="${fn:escapeXml(title)}" var="title"/></c:if>
```
silentsakky, line 31: most of the sanitization for altkey, clazz is done when it's getting used (line no 46/47) and when it's already passed through fmt:message function, any reason to not follow same methodology here?

deepakzimb: I didn't change old changes. we can add fn:escapeXml when a variable is set like title and in this file where ever we are using we can use sanitized value.

deepakzimb: @silentsakky I checked  and apply escapexml on leaf element. If I apply same on line 30,31,32 that is not working and HTML is broken after fix so I applied solution on leaf node.

silentsakky (approved):

zimsuchitgupta (approved):

PR #751, opened 2022-11-28
ZCS-12587, ZCS-12588: added keys for Active in zimlet list and Discard in filter action

k-kato:
Adding the following keys in ZmMsg properties files:
* `activeInZimletsListHeader` (`Active` in English): it is used in Preferences -> Zimlet list 
* `discardMessage` (`Discard` in English): it is used in Add/Edit Filter dialog

The values are the same as `active` and `discard`, respectively. There is no change in the view of Classic UI.
The PR adds the ability to change the value of the keys as a user prefers, as necessary.


ykomiyam (approved):
It looks good.

miteshsavani810 (approved):

silentsakky (approved):

PR #752, opened 2022-11-29
ZCS-12549: add ability to specify favicon directory in a skin

k-kato:
**Enhancement:**
Currently favicon path is hard coded in ZmMailApp.js. Some zimbra users want to change favicon per skin.

**Changes:**
* `FaviconDir` param is added to `skin.properties`. It is commented out by default. Only Network Edition users are allow to change favicon as same as logo image.
* The value of `FaviconDir` is set in `ZmSkin.hints.favicon.dir` as string. It will be 0-length string if `FaviconDir` is not defined in `skin.properties`.
* Update `ZmMailApp.prototype._setFavIcon` to use `skin.hints.favicon.dir` if it has been set.

Note:
The url is sanitized at `appendChild` in `Dwt.setFavIcon`.

### DIFF
--- a/WebRoot/js/zimbraMail/mail/ZmMailApp.js
+++ b/WebRoot/js/zimbraMail/mail/ZmMailApp.js
@@ -2113,12 +2113,19 @@ function() {
 ZmMailApp.prototype._setFavIcon =
 function(unread) {
     var url;
-    if (unread == 0) {
-        url = [appContextPath, "/img/logo/favicon.ico"].join("");
-    } else if (unread > 9) {
-        url = [appContextPath,"/img/logo/favicon_plus.ico"].join("");
+    var dir;
+    if (skin.hints.favicon && skin.hints.favicon.dir) {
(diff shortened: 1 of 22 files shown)


WebRoot/skins/_base/base/skin.properties
```diff
@@ -54,6 +54,7 @@ LogoImgDir		= @AppContextPath@/skins/_base/logos
 #
 #	LogoImgDir		= @AppContextPath@/skins/@SkinName@/logos
 #	LogoURL			= http://www.yourcompany.com
+#	FaviconDir		= @AppContextPath@/skins/@SkinName@/img/favicon
```
silentsakky, line 57: we shouldn't change anything in base and base2 directories as those are old base skins and we are using base3 only currently

k-kato: I reverted changes in base and base2 directories.

silentsakky (approved):

miteshsavani810 (approved):

PR #757, opened 2022-11-30
ZCS-12601: disable popup notification settings on non-supported browsers

k-kato:
**Problem:**
There is a browser which does not support Notifications function like IE11. The behavior is inconsistent around popup notification settings.
* When Preferences -> Mail -> Receiving Messages -> Arrival Notifications -> "Show a popup notification" is unchecked, it is shown on both Chrome and IE11.
* When Preferences -> Mail -> Receiving Messages -> Arrival Notifications -> "Show a popup notification" is checked, it is not shown on IE11.

**Enhancements:**
* When Notifications function is not supported on a browser, the setting should be shown with "disabled" status (unclickable) rather than hidden (invisible)
* In addition, the related setting "Notification Folders" with two radio buttons should also be shown with "disabled" status.
* "Show a popup notification" in Calendar settings should be shown in "disabled" status as well.
* It is necessary that an ability to change the label "Show a popup notification" when the setting is disabled. (e.g. "Show a popup notification (it is not supported on your browser)" )

**Changes:**
* set the following items to "disabled" when notifications functions is not supported on a browser
   * "Show a popup notification" checkbox in Mail setting (MAIL_NOTIFY_TOASTER)
   * "Show a popup notification" checkbox in Calendar setting (CAL_REMINDER_NOTIFY_TOASTER)
   * "Display notifications for new messages in Inbox" and "Display notifications for new messages in any folder" radio buttons (MAIL_NOTIFY_ALL)
* add `ZmMsg.showPopupDisabled` which is used when notifications functions is not supported on a browser
    * The value is the same as `ZmMsg.showPopup` on the PR.  An administrator can set a different message for supported browsers and for non-supported browsers .


k-kato:
IE11 is no longer supported. Close the PR.

PR #761, opened 2022-12-02
ZCS-9175, ZBUG-1876: validation of Email Address field in external account setting

k-kato:
**Problems:**
There are two problems in external imap/pop account setting. `zimbraDataSourceEmailAddress` is not set in a data source object in both cases. It should always be set and unique for check of duplication and for sieve filter execution.

* ZCS-9175
    * When the same email address is set in "Email address" and "Username of Account", `emailAddress` attribute is not set in CreateDetaSourceRequest. 

* ZBUG-1876
    * When "Email address" field is empty, `emailAddress` attribute is not set in CreateDetaSourceRequest.

**Changes:**
* ZCS-9175
    * Add the entered string to datasource object which is being edited

* ZBUG-1876
    * Add empty check on Email address
    * Generate an email address and set it to the list of accounts in Accounts section when it is empty.

### DIFF
--- a/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
+++ b/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
@@ -2756,6 +2756,7 @@ function(evt) {
 	if (!m) return;
 
 	var dataSource = this._currentAccount;
+	dataSource.email = email;
 	if (dataSource.userName == "") {
 		this._setControlValue("USERNAME", section, m[1]);
 	}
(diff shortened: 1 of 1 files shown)


miteshsavani810 (approved):

JayaG (approved):

WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
```diff
@@ -1616,6 +1622,15 @@ function(account, id) {
 	return identity && (typeof prop == "string" ? identity[prop] : identity[prop]());
 };
 
+ZmAccountsPage.prototype.__mandatoryEmail =
+function(str) {
+	if (!str || str.trim().length == 0) {
+		this._errorMsg = AjxStringUtil.htmlEncode(AjxMessageFormat.format(ZmMsg.fieldNameIsARequiredField, [AjxMessageFormat.format(ZmMsg.emailAddr)]));
```
silentsakky, line 1628: we are taking values from translation strings only and not using any user data so do we need htmlEncode?

k-kato: Fixed it. Thank you for the catch.

silentsakky (approved):

PR #765, opened 2022-12-15
ZBUG-2691: set favicon in calendar shown in a separate window

k-kato:
**Issue:**
* favicon is not shown in a calendar shown in a separate window. (right-click a calendar and click "Open in a separate window")

**Changes:**
* In `h/rest`, call `app:skin` first and then call `fmt:setBundle` after a skin is identified. All tag files called from here can use a right ZhMsg. If ZhMsg exists in a skin directory, it is used. If it does not exist, default ZhMsg is used.
* `app:skin` in `restView.tag` is no longer necessary.


silentsakky (approved):

damini-vashishtha (approved):

lavegupta (approved):

k-kato:
I found that the logic needed to be modified again. I will update the PR.

k-kato:
I fixed the issue and updated the description. Please review it again.

silentsakky (approved):

deepakzimb (approved):

PR #771, opened 2022-12-23
ZCS-12637 WCAG Re-opened Issues Part 1

miteshsavani810:
Fixed role and aria label value with User action dropdown.
- Removed username field from keyboard tab navigation as that element is not interactive.
- Updated role and aria-label on User action dropdown element.
See the mail-client changes: https://stash.corp.synacor.com/projects/ZIMBRA/repos/mail-client/pull-requests/340/overview


WebRoot/js/zimbraMail/core/ZmZimbraMail.js
```diff
@@ -2360,12 +2359,15 @@ ZmZimbraMail.DEFAULT_CONTACT_ICON_SMALL = appContextPath + "/img/large/ImgPerson
 ZmZimbraMail.prototype.getDropMenuOptions =
 function(parent, parentElement, adminUrl) {
 
-	var button = new DwtLinkButton({parent: parent, className: DwtButton.LINK_BUTTON_CLASS, parentElement: parentElement, elementTag: "DIV"});
+	var login = appCtxt.getLoggedInUsername();
+	var username = (appCtxt.get(ZmSetting.DISPLAY_NAME)) || login;
```
silentsakky, line 2363:
```suggestion
	var username = appCtxt.get(ZmSetting.DISPLAY_NAME) || login;
```

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
```diff
@@ -2385,17 +2387,20 @@ function(parent, parentElement, adminUrl) {
 
 	if (supportedHelps.indexOf("productHelp") !== -1) {
 		mi = menu.createMenuItem("documentation", {text: ZmMsg.productHelp});
+		mi.getHtmlElement().setAttribute('aria-label', ZmMsg.productHelp + ": link");
```
silentsakky, line 2390: do we need to translate link word?

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
k-kato, line 2363: If `login` is used here only, combine them.
```suggestion
	var username = appCtxt.get(ZmSetting.DISPLAY_NAME) || appCtxt.getLoggedInUsername();
```

k-kato (approved):

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
miteshsavani810: Done.

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
miteshsavani810: Added translation.

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
miteshsavani810: Done.

WebRoot/js/zimbraMail/core/ZmZimbraMail.js
```diff
@@ -2385,17 +2385,20 @@ function(parent, parentElement, adminUrl) {
 
 	if (supportedHelps.indexOf("productHelp") !== -1) {
 		mi = menu.createMenuItem("documentation", {text: ZmMsg.productHelp});
+		mi.getHtmlElement().setAttribute('aria-label', ZmMsg.productHelp + ": " + ZmMsg.link);
```
silentsakky, line 2388: should we merge and create new key? for some languages we might require to change sequence of words which is not possible currently

miteshsavani810: We have intentionally added `aria-label` for NVDA tool with two part as `<option name>: <option type>` such that it will announce `<option name>` and it will pause for 1 second and then it will announce type of that option `<option type>`. As User action dropdown contains few options which are link and as per requirement NVDA tool should announce that option as `Link`

silentsakky (approved):

PR #775, opened 2022-12-28
ZCS-12725: added error handling for skin name

k-kato:
**Issue:**
There is an edge case that zimbraPrefSkin cannot be fetched from auth result in launchZCS.jsp due to an error/exception on backend. In such a case, (part of) error message is unexpectedly set in `skin` variable at the following code.
```
String skin = authResult.getSkin();
```
The value of `skin` is not a valid skin name. When it happens, Classic UI is not shown at all - blank page is shown.

**Fix:**
Added error handling to launchZCS.jsp.
* If skin name cannot be fetched (null) or it is invalid name, set `skin` to an empty string.
* a valid or an empty skin name is passed to `<app:skinAndRedirect defaultSkin="${skin}" />`.  If `${skin}` is empty, a default skin is identified in the tag.
* get a skin identified in the tag using `getAttribute`, and set `skin` to the value.

### DIFF
--- a/WebRoot/public/launchZCS.jsp
+++ b/WebRoot/public/launchZCS.jsp
@@ -1,6 +1,8 @@
 <%@ page buffer="8kb" session="true" autoFlush="true" pageEncoding="UTF-8" contentType="text/html; charset=UTF-8" %>
 <%@ page import="java.util.*,javax.naming.*,com.zimbra.client.ZAuthResult" %>
+<%@ page import="java.util.regex.Pattern" %>
 <%@ page import="com.zimbra.cs.taglib.bean.BeanUtils" %>
+<%@ page import="com.zimbra.common.util.ZimbraLog" %>
 <%@ taglib prefix="zm" uri="com.zimbra.zm" %>
 <%@ taglib prefix="app" uri="com.zimbra.htmlclient" %>
 <%@ taglib prefix="fmt" uri="com.zimbra.i18n" %>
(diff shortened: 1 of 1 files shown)


WebRoot/public/launchZCS.jsp
```diff
@@ -41,11 +43,27 @@
...
+
+    if (skin == null || !Pattern.matches("^[A-Za-z0-9]+$", skin)) {
+        skin = "";
+        ZimbraLog.webclient.warn("a valid skin name could not be fetched. Default skin is used.");
     }
+    pageContext.setAttribute("skin", skin, PageContext.REQUEST_SCOPE);
 %>
 <app:skinAndRedirect defaultSkin="${skin}" />
 <%
+    try {
```
silentsakky, line 61: indentation

k-kato: fixed it

silentsakky (approved):

miteshsavani810 (approved):

lavegupta (approved):

damini-vashishtha (approved):

k-kato:
Even when user can login, Preferences > General > Theme options are broken. If user chooses such a broken option and saves it, an invalid value is set to zimbraPrefSkin. After that, a correct skin is never used. It must be avoided.
The PR does not solve the issue. A root cause needs to be fixed.
Close the PR.

PR #783, opened 2023-01-20
ZCS-12823: WCAG | Classic UI | Add focus to tinymce special character dialog

lavegupta:


miteshsavani810 (approved):

tiny_mce/patches/11-fix-keyboard-focus-in-charmap.diff
```diff
@@ -0,0 +1,96 @@
+Only in tinymce-4.9.11: .DS_Store
```
silentsakky, line 1: we should not commit .DS_Store files

lavegupta: Done

silentsakky (approved):

HimanshuSojitraSynacor (approved):

PR #786, opened 2023-01-27
ZCS-12542: increase the width of address picker dialog when Sharing feature and GAL are disabled

k-kato:
**Issue:**
When zimbraFeatureSharingEnabled and zimbraFeatureGalEnabled are FALSE, the width of address picker dialog is narrow.

**Root cause:**
When they are FALSE, "Show names from: + Options" is not included in the dialog. It is not just invisible but html elements for the label and option does not exist. Then the width of the dialog decreases.

**Fix:**
* set the width of the dialog to `900px` when the "Show names from:" option does not exist
* modified a template
    * When the option exists, it works as it is.
    * When it does not exist, `td` element with `width='300px'` is used.

**Additional information:**
When the attributes are TRUE, the width of the dialog depends on the max string length of "Show name from" options.
For example, with the following options,
(English)
* Global Address List
* Personal and Shared Contacts
* Contacts

(Japanese) 
* グローバルアドレスリスト
* 個人連絡先と共有連絡先
* 連絡先

the width of the dialog is set as follows:
(English):
* both TRUE:  927px
* both FALSE: 621px  -> 900px after the fix

(Japanese)
* enabled:  880px
* disabled: 622px  -> 900px after the fix

### DIFF
--- a/WebRoot/templates/abook/Contacts.template
+++ b/WebRoot/templates/abook/Contacts.template
@@ -985,7 +985,7 @@
 				<td>
 					<table role="presentation" width='100%'>
 						<tr>
-							<td width='600'>
+							<td width='600px'>
 								<table role="presentation" class='ZPropertySheet' cellspacing='6' id='${id}_searchTable'>
 									<$ if (data.detailed) { $>
 									<tr id='${id}_searchNameRow'>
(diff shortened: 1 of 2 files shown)


WebRoot/templates/abook/Contacts.template
```diff
@@ -1025,6 +1025,13 @@
 											<td id='${id}_listSelect'></td>
 										</tr>
 									</table>
+								<$ } else { $>
+									<table role="presentation" class='ZPropertySheet' cellspacing='6' style="visibility: hidden;">
+										<tr>
+											<td class='Label nobreak' id='${id}_listSelectLbl_invisible'><$= ZmMsg.showNames $></td>
+											<td id='${id}_listSelect_invisible'><$= ZmMsg.searchPersonalSharedContacts $></td>
+										</tr>
+									</table>
```
k-kato, line 1034: * `style="visibility: hidden;"` is set for the table.
* comparing with code at lines 1022-1027, different classname is set on each so that the change will not affects any js code which getting an element of `_listSelectLbl` or `_listSelect`
    * `_listSelectLbl` -> `_listSelectLbl_invisible`
    * `_listSelect` -> `_listSelect_invisible`

miteshsavani810 (approved):

JayaG (approved):

silentsakky:
Fix looks unnecessarily complicated, we can fix it with a single line change
ZmContactPicker.js line no 512
`this.setSize("800px");`
In the current code we are not passing px in the string and hence it was not getting appended in dom, and hence size was not getting reflected

k-kato:
Thanks @silentsakky, I didn't know the reason why the setSize didn't work. It works with `px` as you mentioned.
I changed it to `900px` and fixed Contacts.template. The description of the PR has also been updated.

silentsakky (approved):

PR #769, opened 2022-12-20
ZCS-12677: add ability to load css in a skin on print page

k-kato:
**Problem:**
css file in a skin directory is not loaded on print message/contact/appointment/task page.

**Changes:**
* `head.tag`
    * specify `print` in `client` param for print page
    * add `skin.css` for print page
* `WebRoot/skins/_base/base3/print.css`: add empty file. Administrator can add style(s) as necessary.
* `manifest.xml` in each skin directory
    * add `<print>` element which contains the `print.css`
    * move css file(s) in `<common>` element to `<advanced>` and `<standard>` so that the css file(s) is not loaded on print page. Otherwise, style/layout of print page will be changed unexpectedly.
    Note: files in `<common>` are always loaded on any client type - advanced, standard and print
    * indentation: use tab space
    * line break: use LF on all skins

PR #788, opened 2023-01-30
ZCS-12677: add ability to load css in a skin on print page

k-kato:
**Problem:**
css file in a skin directory is not loaded on print message/contact/appointment/task page.

**Changes:**
* `head.tag`
    * specify `print` in `client` param for print page
    * add `skin.css` for print page
* `WebRoot/skins/_base/base3/print.css`: add empty file. Administrator can add style(s) as necessary.
* `manifest.xml` in each skin directory
    * add `<print>` element which contains the `print.css`
    * add `<cssIgnore>` element not to include css files defined in `common / css` section.
        * see https://github.com/Zimbra/zm-ajax/pull/116
    * indentation: use tab space
    * line break: use LF on all skins

FYI, you can ignore changes of indentation (white space / tab space) and line breaks (CRLF / LF) in code diff page.
File changed -> Gear icon -> Check "Hide whitespace" -> Click "Apply and reload"

### DIFF
--- a/WebRoot/skins/bare/manifest.xml
+++ b/WebRoot/skins/bare/manifest.xml
@@ -44,14 +44,14 @@
 			<file>../../css/zhtml.css</file>
 			<file>../_base/base3/zhtml.css</file>
 		</css>
-	</standard>		
-    <print>
-        <css>
-            <file>../_base/base3/print.css</file>
-        </css>
-        <cssIgnore>
-            <file>../_base/base3/skin.css</file>
-            <file>skin.css</file>
-        </cssIgnore>
-    </print>
+	</standard>
+	<print>
+		<css>
+			<file>../_base/base3/print.css</file>
+		</css>
+		<cssIgnore>
+			<file>../_base/base3/skin.css</file>
+			<file>skin.css</file>
+		</cssIgnore>
+	</print>
 </skin>
--- a/WebRoot/WEB-INF/tags/head.tag
+++ b/WebRoot/WEB-INF/tags/head.tag
@@ -41,13 +41,15 @@
     <c:choose>
         <c:when test="${not print}">
             <c:set var='css' value='/css/common,login,images,skin.css'/>
+            <c:set var="clientType" value="standard"/>
         </c:when>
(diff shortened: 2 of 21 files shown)


miteshsavani810 (approved):

WebRoot/skins/bare/manifest.xml
```diff
@@ -44,5 +44,14 @@
 			<file>../../css/zhtml.css</file>
 			<file>../_base/base3/zhtml.css</file>
 		</css>
-	</standard>		
+	</standard>
+	<print>
+		<css>
+			<file>../_base/base3/print.css</file>
+		</css>
+		<cssIgnore>
```
silentsakky, line 52: we are including skin.css in https://github.com/Zimbra/zm-web-client/pull/788/files#diff-d34b0fa9ff6036a11669d2a34ee3404cc6853a6b851e5932af54a0610d7e2ae6R47 and then ignoring here, looks a bit weird

WebRoot/skins/bare/manifest.xml
```diff
@@ -44,5 +44,14 @@
 			<file>../../css/zhtml.css</file>
 			<file>../_base/base3/zhtml.css</file>
 		</css>
-	</standard>		
+	</standard>
+	<print>
+		<css>
+			<file>../_base/base3/print.css</file>
```
silentsakky, line 50: can't we load this as part of the standard block? line no 44

WebRoot/skins/bare/manifest.xml
k-kato: Previously I tried another approach not to load `../_base/base3/skin.css` when `client=print` is specified.
* https://github.com/Zimbra/zm-web-client/pull/769

However, it broke another pages. There are some pages which request skin.css without specifying client type. In this case, only `common` block is checked. No css was loaded with PR 769.

We need to keep `common` block as it is to avoid regression on other pages.
The purpose of the change is to load only print.css in a skin. So we need to remove css file(s) added in `common` block.

JayaG (approved):

PR #789, opened 2023-01-30
ZBUG-3072: German Language/WebUI: Error when snoozing an appointment reminder with "Repeat at time of event" option.

lavegupta:


WebRoot/js/zimbraMail/calendar/view/ZmApptQuickAddDialog.js
```diff
@@ -338,11 +338,10 @@ function() {
 		var option = ZmApptViewHelper.REPEAT_OPTIONS[i];
 		this._repeatSelect.addOption(option.label, option.selected, option.value);
 	}
-
 	//reminder DwtSelect
     var	displayOptions = [
 		ZmMsg.apptRemindNever,
-        ZmMsg.apptRemindAtEventTime,
```
silentsakky, line 345: looks like we have quite a bit of code duplication between ZmMsg and AjxMsg, I thought it was only for one key
any idea when this was introduced and which one of both is widely used? so we will preserve widely used ones and remove lesser used one

PR #791, opened 2023-02-01
ZCS-12917: changed to remove ZM_TEST and ZM_LOGIN_CSRF cookies first

k-kato:
**Issue:**
When zimbraMailURL is changed from `/` to `/test/path`, for example, and user accesses `https://host/` (without the path), login fails with an error "An unknown error has occurred."
Steps:
1. set zimbraMailURL to `/test/path`
2. access `https://host/`
3. login to Classic UI. It succeeds at this time. The URL changes to `https://host/test/path/?client=advanced`
4. log out.  The URL is `https://host/test/path/?loginOp=logout`
5. access `https://host/` without the path
6. try to login, but it fails.

Easier way:
1. set zimbraMailURL to `/test/path`
2. access `https://host/test/path`
3. access `https://host/`
4. try to login, but it fails.

**Root cause:**
When login page accessed, `ZM_LOGIN_CSRF` is created on a browser. The path of the cookie depends on the access URL.
* `https://host` -> the path of the cookie is `/`
* `https://host/test/path` -> the path of the cookie is `/test/path`

When a browser has the both cookies, login from `https://host` fails because `(param.login_csrf eq cookie.ZM_LOGIN_CSRF.value)` returns `false`. `cookie.ZM_LOGIN_CSRF` refers a cookie for `/test/path`, not for `/`. Then it falls back to `c:otherwise` block.
```
at line 145
<c:when test="${(not empty param.login_csrf) && (param.login_csrf eq cookie.ZM_LOGIN_CSRF.value)}">
   // zm:login process
</c:when>
<c:otherwise>
	<!-- on failure of csrf show error to user -->
	<c:set var="errorCode" value="unknownError"/>
	<fmt:message var="errorMessage" key="unknownError"/>
</c:otherwise>
```

**Fix:**
* Remove ZM_LOGIN_CSRF cookies for both `/` and `/test/path` before updating it.
* Do the same for ZM_TEST cookies


WebRoot/public/login.jsp
```diff
@@ -384,6 +385,29 @@ if (application.getInitParameter("offlineMode") != null) {
 </c:url>
 
 <%
+	Cookie testCookieRootPath = new Cookie("ZM_TEST", "true");
+	testCookieRootPath.setMaxAge(0);
+	testCookieRootPath.setPath("/");
+	response.addCookie(testCookieRootPath);
```
silentsakky, line 391: this is already done at line no 413, are we repeating it?

WebRoot/public/login.jsp
```diff
@@ -384,6 +385,29 @@ if (application.getInitParameter("offlineMode") != null) {
...
+
+	Cookie csrfCookieRootPath = new Cookie("ZM_LOGIN_CSRF", "");
+	csrfCookieRootPath.setMaxAge(0);
+	csrfCookieRootPath.setPath("/");
+	response.addCookie(csrfCookieRootPath);
+
+	String zimbraMailURL = (String) pageContext.getAttribute("zimbraMailURL");
```
silentsakky, line 398: instead of this entire code block, can't we just put the below statements at lines no 418 and 412?
```
String zimbraMailURL = (String) pageContext.getAttribute("zimbraMailURL");
testCookieNonRootPath.setPath(zimbraMailURL);
```

WebRoot/public/login.jsp
k-kato: Before the fix, for example,
1. access `https://host`. A cookie for `/` is created.
2. access `https://host/test/path`. A cookie for `/test/path` is created.
3. access `https://host`. The cookie for `/` is updated. The cookie for `/test/path` still remains and the value is not updated. 
4. access `https://host/test/path`. The cookie for `/test/path` is update. The cookie for `/` still remains and the value is not updated. 

The unnecessary cookie whose path is different from the access path needs to be removed.
On JSP side, `request.getContext` or `getRequestURI` always returns `/test/path`. Then the jsp cannot identify the access URL specified on the browser.
So the fix is to remove both cookies first before line 413.

WebRoot/public/login.jsp
k-kato: Even when zimbraMailURL is set to `/test/path/`, user can access `https://host/` without the path. As described in another comment, jsp cannot identify which cookie of a path needs to be removed / updated. So the code you suggested will not solve the issue.

WebRoot/public/login.jsp
silentsakky: not sure why we are rendering login page when accessing https://host and zimbraMailURL is set, nginx should send a redirect to correct address, this way we don't have to handle these scenarios in login.jsp

k-kato: I get your point. I will update it.

k-kato: Updated the PR.
The issue is solved by specify the path `/` for ZM_LOGIN_CSRF and ZM_TEST cookies. The cookies will work on both `https://host/` and `https://host/test/path`.

silentsakky (approved):

k-kato:
I found that ZM_LOGIN_CSRF cookie was not removed after login at line 154-156. I will fix it and update the PR.

k-kato:
I added a commit.
@silentsakky please review it again.

silentsakky (approved):

miteshsavani810 (approved):

PR #793, opened 2023-02-01
ZCS-12907: set context path on password recovery page

k-kato:
**Issue:**
Context path was not set in Password Recovery page. 

**Fix:**
* add context path to the path of jsp and img files.


k-kato:
found that path of "Back to Sign in" also need to be fixed. I will fix it and update the PR.

k-kato:
updated

silentsakky (approved):

miteshsavani810 (approved):

PR #578, opened 2020-03-09
PREAPPS-4347 :  Reskin of error.jsp and remove of url from 2fa and forgetpassword flow

auminpatel:
reskin of error.jsp page and removing zimbra url from 2fa and forgort-password flow and url for the login page can be cusotomised in WebRoot/skins/harmony/skin.properties

PR #795, opened 2023-02-03
ZCS-12944: changed to get skin from pageContext on error.jsp

k-kato:
**Issue:**
When error.jsp is shown, default skin harmony is always used.
When user accesses ZWC with virtual hostname and zimbraPrefSkin is set on the domain, css files defined in the skin should be used.

**Fix:**
`<app:skinAndRedirect />` at line 26 gets a skin of a domain based on access URL. The skin is set in "skin" attribute of pageContext.
After the fix, skin is identified as follows:
1. If request url for error.jsp has "skin" parameter, it is used.
2. If not, a skin set in pageContext is used.
3. If both empty or null, zimbraDefaultSkin in init parameter is used.

The PR is adding the 2nd logic.

### DIFF
--- a/WebRoot/public/error.jsp
+++ b/WebRoot/public/error.jsp
@@ -29,7 +29,11 @@
 
 <%
 	String skin = request.getParameter("skin");
-	if (skin == null || !Pattern.matches("^[0-9A-Za-z]+$", skin)) {
+	if (skin == null) {
+		Object o = pageContext.getAttribute("skin", PageContext.REQUEST_SCOPE);
+		skin = String.valueOf(o);
+	}
+	if (skin == null || "null".equals(skin) || !Pattern.matches("^[0-9A-Za-z]+$", skin)) {
 		skin = application.getInitParameter("zimbraDefaultSkin");
 	}
 %>


WebRoot/public/error.jsp
```diff
@@ -29,7 +29,11 @@
 
 <%
 	String skin = request.getParameter("skin");
-	if (skin == null || !Pattern.matches("^[0-9A-Za-z]+$", skin)) {
+	if (skin == null) {
+		Object o = pageContext.getAttribute("skin", PageContext.REQUEST_SCOPE);
+		skin = String.valueOf(o);
+	}
+	if (skin == null || "null".equals(skin) || !Pattern.matches("^[0-9A-Za-z]+$", skin)) {
 		skin = application.getInitParameter("zimbraDefaultSkin");
```
silentsakky, line 37: I am a little bit confused on why we need this here, skinAndRedirect is already handling logic of getting correct skin parameter so why do we need to get zimbraDefaultSkin here?

k-kato: The logic has been added on https://github.com/Zimbra/zm-web-client/pull/578
In `error.jsp`, a skin can be specified using a url parameter like `https://host/aaa?skin=xxx`
If `skin` parameter is specified but the value does not match `^[0-9A-Za-z]+$`, `zimbraDefaultSkin` is set as a skin here.

The 2nd logic being added in the PR (which is described in Description of the PR) is executed only when a request url does not have `skin` parameter.

miteshsavani810 (approved):

JayaG (approved):

silentsakky:
I am trying out some alternative methods to solve this as I feel we are overcomplicating things currently, so wait for sometime on this

k-kato:
Created PRs for a different approach.
* https://github.com/Zimbra/zm-web-client/pull/803
* https://github.com/Zimbra/zm-taglib/pull/47

Close the PR.

PR #801, opened 2023-02-16
ZCS-13042: fixed for device with small screen resolution

k-kato:
**Issue:**
Looping access occurs when accessing Classic UI from PC with small resolution.

Steps to reproduce:
1. set account's zimbraPrefClientType to "advanced"
2. access Login page using a PC browser. The display resolution is less than 600x800 (width x height)
3. choose "Classic" client type and log in to the account. The URL has "client=advanced" parameter. (https://host/?client=advanced )
6. remove the parameter and access https://host/
7. Looping access occurs. The URL has "client=standard" parameter.

**Changes:**
* launchZCS.jsp: changed the parameter to `screenSize=small` for redirection to login.jsp
* login.jsp
    * when parameter has `screenSize=small` and Modern UI is supported, set `client` to `modern`
    * when parameter has `screenSize=small` and Modern UI is not supported, set `client` to `notfound` because both advanced and modern are unavailable for the device. In addition, an error message is set.
    * If `client` is `notfound`, force to logout. If the error message has been set for small screen device, it is shown in login page.

Related PR:
* https://github.com/Zimbra/zm-ajax/pull/121

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -43,7 +43,7 @@
 <fmt:setBundle basename="/messages/ZMsg" var="zmsg" scope="request"/>
 
 <%-- query params to ignore when constructing form port url or redirect url --%>
-<c:set var="ignoredQueryParams" value=",loginOp,loginNewPassword,totpcode,loginConfirmNewPassword,loginErrorCode,username,email,password,zrememberme,ztrusteddevice,zlastserver,client,login_csrf,"/>
+<c:set var="ignoredQueryParams" value=",loginOp,loginNewPassword,totpcode,loginConfirmNewPassword,loginErrorCode,username,email,password,zrememberme,ztrusteddevice,zlastserver,client,login_csrf,screenSize,"/>
 
 <%-- get useragent --%>
 <zm:getUserAgent var="ua" session="false"/>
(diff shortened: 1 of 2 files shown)


sitaram-mulik-synacor (approved):

WebRoot/public/login.jsp
```diff
@@ -256,7 +256,20 @@
                         
                         <c:set var="isUpgradedMailbox" value="${isUpgradedMailbox}" />
                         <c:set var="prefClientType" value="${requestScope.authResult.prefs.zimbraPrefClientType[0]}" />
-                        
+
+                        <c:if test="${param.screenSize eq 'small' && empty client}">
```
miteshsavani810, line 260: @k-kato 
What happen if user logged in the classic client and then reduce the screen size and reload the browser.
Will it switch to modern client ?

k-kato: @miteshsavani810 It depends on conditions.

(A) When zimbraPrefClientType is advanced and Default client is selected in login page, `client=advanced` is not added to URL parameter. In this case, when screen size is reduced and the page is reloaded, it is redirected to `https://host/?screenSize=small` to access login page. If Modern UI is supported (`/modern` exists or the browser is not IE11), it is redirected to Modern client. If not, the user logs out and an error is shown in login page.

(B) When Classic client is selected in login page, `client=advanced` is added to URL parameter regardless of zimbraPrefClientType. In this case, when screen size is reduced and the page is reloaded, it is not redirected. Classic UI is still shown. It is processed by `<c:if test="${param.client ne 'advanced'}">` in launchZCS.jsp at line 330.

k-kato: This means, the redirection does not happen when user selects Classic in login page explicitly.
Then the error message includes `If you still want to use it, please choose "Classic" in Web App Version.` as defined in https://github.com/Zimbra/zm-ajax/pull/121.

miteshsavani810: Ok. Got it. Thanks.

miteshsavani810 (approved):

PR #816, opened 2023-05-10
Email doesn't show up immediately in the Inbox zimbra

mohsenmkh1993:
I dont know the reason why thr contents of emails are not display in the zimbra mailbox?


silentsakky:
Not sure why this PR is opened, closing it.

PR #819, opened 2023-05-18
PREAPPS-7295: Disable modern web app

lavegupta:


WebRoot/public/login.jsp
```diff
@@ -90,18 +90,24 @@
 
 <%
     // check if modern package exists
-    Boolean modernSupported = (Boolean) application.getAttribute("modernSupported");
-    if(modernSupported == null) {
+    Boolean isIndexFileExist = (Boolean) application.getAttribute("isIndexFileExist");
```
silentsakky, line 93:
```suggestion
    Boolean indexFileExists = (Boolean) application.getAttribute("indexFileExists");
```

WebRoot/public/login.jsp
```diff
@@ -90,18 +90,24 @@
...
         }
-        application.setAttribute("modernSupported", modernSupported);
+        application.setAttribute("isIndexFileExist", isIndexFileExist);
     }
 %>
-<c:set var="modernSupported" value="<%=modernSupported%>" />
+
+<c:set var="isIndexFileExist" value="<%=isIndexFileExist%>" />
+
+<zm:getDomainInfo var="domainInfo" by="virtualHostname" value="${zm:getServerName(pageContext)}"/>
```
silentsakky, line 107: on line 21 also we are doing same call, we should see if we can merge it

WebRoot/js/zimbraMail/share/model/ZmSettings.js
```diff
@@ -892,6 +892,8 @@ function() {
 	this.registerSetting("PUBLIC_URL",						{type:ZmSetting.T_CONFIG});
 	this.registerSetting("ADMIN_URL",						{type:ZmSetting.T_CONFIG});
 	this.registerSetting("DISABLE_SENSITIVE_ZIMLETS_IN_MIXED_MODE",		{type:ZmSetting.T_CONFIG});
+	this.registerSetting("DISABLE_MODERN_CLIENT",			{name:"zimbraModernWebClientDisabled", type:ZmSetting.T_DOMAIN, dataType:ZmSetting.D_BOOLEAN, defaultValue:false});
```
k-kato, line 895: A name of an attribute is usually `...Enabled` for boolean type. Is there any reason not to define `zimbraModernWebClientEnabled` with default value `TRUE`?

k-kato:
[Dockerfile](https://github.com/Zimbra/zm-web-client/blob/develop/Dockerfile) needs to be updated as well.

WebRoot/js/zimbraMail/share/model/ZmSettings.js
silentsakky: yes we can do that, only concern with that is if somehow ldap attribute is not present then by default this logic will show modern ui option but if we change it to Enabled then by default it will hide the option

k-kato: I got it. Let's keep it as it is.

lavegupta: Changed it to **zimbraModernWebClientEnabled** and made default value **TRUE**

WebRoot/public/login.jsp
lavegupta: Done

WebRoot/public/login.jsp
```diff
@@ -90,18 +90,23 @@
...
-        application.setAttribute("modernSupported", modernSupported);
+        application.setAttribute("indexFileExists", indexFileExists);
     }
 %>
-<c:set var="modernSupported" value="<%=modernSupported%>" />
+
+<c:set var="indexFileExists" value="<%=indexFileExists%>" />
+
+<c:set var="modernWebClientEnabled" value="${domainInfo.attrs.zimbraModernWebClientEnabled}" />
+<c:set var="modernSupported" value="${(modernWebClientEnabled) and (indexFileExists)}" />
```
silentsakky, line 108:
```suggestion
<c:set var="modernSupported" value="${modernWebClientEnabled and indexFileExists}" />
```

miteshsavani810 (approved):

WebRoot/public/login.jsp
lavegupta: Done

silentsakky (approved):

zimsuchitgupta (approved):

PR #828, opened 2023-07-07
ZCS-2733 remove product help page link if help pages are not available in the build

sitaram-mulik-synacor:
Product help pages will be excluded from next build and patch releases. Those can be include back by installing "zimbra-help" package. This code removes the "Product Help" link from the UI when the help pages are unavailable.

### DIFF
--- a/WebRoot/public/launchZCS.jsp
+++ b/WebRoot/public/launchZCS.jsp
@@ -64,7 +64,7 @@
 	try {
 		productHelpSupported = new java.io.File(application.getRealPath("/help")).exists();
 	} catch (Exception ignored) {
-		// Just in case there's anException, catch it.
+		// Just in case there's an exception, catch it.
 	}
 	application.setAttribute("productHelpSupported", productHelpSupported);
 %>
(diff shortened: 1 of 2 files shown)


silentsakky (approved):

WebRoot/public/launchZCS.jsp
```diff
@@ -58,6 +58,17 @@
 	// Prevent IE from ever going into compatibility/quirks mode.
 	response.setHeader("X-UA-Compatible", "IE=edge");
 %>
+<%
+    // check if help directory exists
+    boolean productHelpSupported = true;
+	try {
+		productHelpSupported = new java.io.File(application.getRealPath("/help")).exists();
+	} catch (Exception ignored) {
+		// Just in case there's anException, catch it.
```
k-kato, line 67:
```suggestion
		// Just in case there's an Exception, catch it.
```
or 
```suggestion
		// Just in case there's an exception, catch it.
```

k-kato (approved):

WebRoot/public/launchZCS.jsp
sitaram-mulik-synacor: Updated the comment, please review @k-kato .

k-kato (approved):

PR #829, opened 2023-07-12
ZBUG-3198 Print Preview comes to new line in Chrome browser.

damini-vashishtha:


WebRoot/css/zhtml.css
```diff
@@ -492,6 +492,10 @@ A:visited { @Text-link@ }
 	padding: 10px;
 }
 
+.zPrintMsgs .MsgBody-html * {
+	page: auto;
+}
+
```
k-kato, line 498: Styles of `@page` defined in an original message will not be applied as side effect, correct?

k-kato: In many case, `@page` style seems to be added by (some?) software/service automatically. The style is applied at print only.
As we discussed, a print page should be printed as we see on a browser. So it is okay to ignore `@page` style by using the `page: auto;` defined here.
I close my question.

k-kato (approved):

miteshsavani810 (approved):

lavegupta (approved):

PR #839, opened 2023-09-22
ZCS-13311: shared items not visible for external users issue

sitaram-mulik-synacor:


sitaram-mulik-synacor:
Automation workflow - https://app.circleci.com/pipelines/github/ZimbraOS/zm-frontend-automation/1020/workflows/2b5a3d57-2f90-43ce-a65f-4cf5fba86476 
(Please refer zimbra10 only, the failing test cases are not releted to this code change and present in the master branch as well)

miteshsavani810 (approved):

HimanshuSojitraSynacor (approved):

WebRoot/js/zimbraMail/mail/ZmMailApp.js
```diff
@@ -1657,7 +1657,8 @@ function(params, callback) {
 
     if (appCtxt.isExternalAccount()) {
         var loadCallback = this._handleLoadLaunch.bind(this, params, callback);
-	    AjxDispatcher.require(["MailCore", "Mail", "Startup2"], true, loadCallback, null, true);
+	    AjxDispatcher.require(["MailCore", "Mail"], true, loadCallback, null, true);
+	    AjxDispatcher.require("Startup2", false, loadCallback, null, true);
```
k-kato, line 1661: Is it okay to call them in different sync/async mode? The former is called with `async true` and the latter with `async false`. There might be a case that Startup2 is loaded faster than the others for some reason. Do they need to be loaded in the specific order; MailCore, Mail and then Startup2?

In addtion, if the issue depends on the order (or timing) of loading modules and other modules like calendar need to be loaded after the Startup2, does `async false` need to be specified on both so that MailCore, Mail and Startup2 are surely loaded here?

sitaram-mulik-synacor: Yes, in fact what I find out during debugging that the Startup2 should be loaded earlier because it has multiple utility functions which are being used during the page load. In case of regular user login scenario, it is being loaded synchronously. However for the external user it was being loaded asynchronously. And I didnt find any need of loading "MailCore" and "Mail" packages asynchronously because they are not needed during the page load. Hence I have kept them as it is.

k-kato: I understood. Thanks Sitaram.

k-kato (approved):

PR #842, opened 2023-10-06
ZBUG-3617: User with & in username is unable to login to webmail

sakshirawat0247:
**RCA**: Somewhere username is getting encoded when sending auth request, most probably the username which is being provided by the username is getting encoded using escapeXml and the trimmedUsername is used to validate the password rules which is also in an encoded form.
**Solution**: encoded form of trimmedUsername is stored in encodedUserName and the password is stored in encodedPwd 
 which is being used to validate the zimbraFeatureAllowUsernameInPassword rule. Removed the escapeXml of JSTL from the trimmedUsername. It will send the unencoded data to hit the API.

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -52,12 +52,13 @@
                                                         || (ua.isOsAndroid)
                                                         || (ua.isIos))}"/>
 <c:set var="totpAuthRequired" value="false"/>
-<c:set var="trimmedUserName" value="${fn:trim(fn:escapeXml(param.username))}"/>
+<c:set var="trimmedUserName" value="${fn:trim(param.username)}"/>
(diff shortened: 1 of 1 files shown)


silentsakky (approved):

k-kato (approved):

miteshsavani810 (approved):

JayaG (approved):

silentsakky:
Please add automation for this change

sakshirawat0247:
> Please add automation for this change

for this feature test case have already been added C7769, C15983, C14868, C14815 earlier in https://github.com/Zimbra/zm-frontend-automation. I can execute these test case again and attach the snapshot of the same

silentsakky (approved):

PR #848, opened 2023-12-18
ZCS-14070, ZCS-14071, ZCS-14073: updated TFA login and setup logic to support email method

k-kato:
Updating two-factor authentication setup and login process to support email method.
Target branch is `feature/2FAEmailCode`.

**Main changes:**
* show "Setup/Disable two-step authentication" link for each method separately in Preferences. Allowed TFA method(s) is defined in zimbraTwoFactorAuthMethodAllowed by administrator.
* regarding email method, zimbraPrefPasswordRecoveryAddress for reset password function is used on TFA as well.
   * when zimbraFeatureResetPasswordStatus is enabled, user must configure and verify zimbraPrefPasswordRecoveryAddress before the user configure TFA email method.
   * when zimbraFeatureResetPasswordStatus is enabled and recovery address is reset, TFA email method becomes disabled. See also Note below.
* when zimbraFeatureTwoFactorAuthRequired is true and any TFA method has not been enabled, TFA setup page is shown after user logs in with username and password. Choose Method page is shown if two (or more) methods are allowed.
* for backward compatibility,
   * if zimbraTwoFactorAuthMethodAllowed is empty, it is regarded as app
   * if zimbraTwoFactorAuthEnabled is TRUE and zimbraTwoFactorAuthMethodEnabled is empty, the latter is regarded as app.

**Note:**
When TFA is required (mandatory), email method has been configured but app method has not been configured, recovery address cannot be changed because it is used for TFA and TFA email method cannot be disabled. We will discuss it later.

Related PRs:
* https://github.com/Zimbra/zm-ajax/pull/128
* https://github.com/Zimbra/zm-taglib/pull/52
* https://github.com/Zimbra/zm-mailbox/pull/1541


WebRoot/js/zimbraMail/share/view/dialog/ZmTwoFactorSetupDialog.js
```diff
@@ -362,18 +672,67 @@ function() {
...
+ZmTwoFactorSetupDialog._GetInfoCallback =
+function(params, disableTFA, method, result) {
+	var dialog = appCtxt.getMsgDialog();
+	dialog.setMessage(ZmMsg.twoFactorAuthGetInfoFailed, DwtMessageDialog.WARNING_STYLE);
+	if (!result || result.isException()) {
+		dialog.popup();
+		return;
+	}
+	try {
+		var response = result.getResponse().Body.GetInfoResponse;
```
silentsakky, line 711: we are already initializing all settings based on GetInfo call at startup check settings.setUserSettings
so do we need another call here?

lavegupta (approved):

WebRoot/js/zimbraMail/share/view/dialog/ZmTwoFactorSetupDialog.js
```diff
@@ -287,53 +579,51 @@ function(currentDivId, result) {
...
-			this._handleTwoFactorAuthSuccess(currentDivId);
-			return;
-		}
-		else if (scratchCodes && scratchCodes[0] && scratchCodes[0].scratchCode) {
-			if (typeof appCtxt !== "undefined") {
-				//Only the server will set ZmSetting.TWO_FACTOR_AUTH_ENABLED. Don’t try to save the setting from the UI.
-				appCtxt.set(ZmSetting.TWO_FACTOR_AUTH_ENABLED, true, false, false, true);
+
+		if (typeof appCtxt !== "undefined") {
+			// scratchCodes is returned only when a first method is enabled.
```
k-kato, line 584: The comment is wrong. Scratch codes are returned when a 2nd method is enabled as well. They are the existing codes.
i.e. they are the same as the ones generated at a 1st method is enabled. If some of them have already been used, the remaining codes are returned.
I will remove the comment.

WebRoot/js/zimbraMail/share/view/dialog/ZmTwoFactorSetupDialog.js
k-kato: yes, regarding zimbraTwoFactorAuthMethodEnabled, zimbraPrefPrimaryTwoFactorAuthMethod and zimbraPrefPasswordRecoveryAddressStatus, they are changed depending on the current configuration.
The logic is a bit complicated for backward compatibility and for consistency between TFA setting and recovery account setting.
The logic may be updated for some reason. (e.g. new method is introduced in the future, any issue is found, etc.)
It is better to get the updated settings using GetInfoRequest/Response for consistency between UI and server, and to avoid duplicate work on server and UI changes.

WebRoot/js/zimbraMail/share/view/dialog/ZmTwoFactorSetupDialog.js
k-kato: done

miteshsavani810 (approved):

k-kato:
Fixed tag names as per changes on zm-taglib.

WebRoot/css/login.css
```diff
@@ -90,6 +90,30 @@ select#client:focus {
 	float:none;
 }
 
+.chooseTFAMethod,
+.chooseTFAMethodBack {
+	float: inline-end;
```
silentsakky, line 95: this is supported in mostly new browsers only https://developer.mozilla.org/en-US/docs/Web/CSS/float
can this give issue in older browsers?

silentsakky (approved):
Only one review comment, rest is good

WebRoot/css/login.css
k-kato: Thank you @silentsakky for pointing it out. I confirmed the style was not applied on IE11 (MS Edge, IE mode). It will be addressed in another ticket to check/update messages and styles.

PR #849, opened 2023-12-22
PREAPPS-7606: Date Filter in Advance setting displays incorrect date if selected date is appended with current time.

sumitTelus:
**Issue in current implementation:** 
Date Filter in Advance setting displays incorrect date if selected date is appended with current time.

**what is observed:**
Current date gets appended to the selected date and its equivalent timeStamp is calculated and sent to API. Since current time is appended it displays incorrect date when server timezone and client timezone varies and it dose not handle DST time. 

**What should it be changed to:**
Removing the time part and sending only date part in long format to the API.

**Example:**
If the selected date is 1/1/2024 then its equivalent timestamp value 1704090386 is sent to ModifyFilterRulesRequest.
Whose equivalent date value is Monday, 1 January 2024 01:26:26 [GMT-05:00]
But in response we get 1704067200. i.e, Sunday, 31 December 2023 19:00:00 [GMT-05:00]
Which is one day before the selected date.

**Solution / Approach:**
Removing the time part and sending only date in long format displays proper date. The timezone calculation is done in backend side considering LDAP attribute for timezone.
For Ex: If selected date is 1/1/2024 then its equivalent timestamp value 1704047400 is sent to ModifyFilterRulesRequest.
Whose equivalent date value is 2024-01-01 00:00:00
and in response we get the same value i.e, 1704047400

### DIFF
--- a/WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
+++ b/WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
@@ -1389,11 +1389,7 @@ function(ev) {
 		if (msg = this._checkCondition(c)) {
 			break;
 		} else {
-			if (c.testType === ZmFilterRule.C_TEST_MAP[ZmFilterRule.C_DATE]) {
-				rule.addCondition(c.testType, c.comparator, Number(c.value), c.subjectMod, c.caseSensitive);
-			} else {
-				rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
-			}
+			rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
 		}
 	}
 	if (!msg) {
--- a/WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
+++ b/WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
@@ -1389,7 +1389,11 @@ function(ev) {
 		if (msg = this._checkCondition(c)) {
 			break;
 		} else {
-			rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
+			if (c.testType === ZmFilterRule.C_TEST_MAP[ZmFilterRule.C_DATE]) {
+				rule.addCondition(c.testType, c.comparator, Number(c.value), c.subjectMod, c.caseSensitive);
+			} else {
+				rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
+			}
 		}
 	}
 	if (!msg) {
@@ -1550,7 +1554,7 @@ function(inputs, conf, field) {
 		if (!date) {
 			return null;
 		}
-		return String(date.getTime() / 1000);
+		return String(new Date(date.getFullYear(),date.getMonth(), date.getDate()).getTime()/1000);
 	}
 	if (type == ZmFilterRule.TYPE_FOLDER_PICKER) {
 		return inputs[field].dwtObj.getData(ZmFilterRuleDialog.DATA);


WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
```diff
@@ -1550,7 +1550,9 @@ function(inputs, conf, field) {
 		if (!date) {
 			return null;
 		}
-		return String(date.getTime() / 1000);
+		AjxDateUtil._serverDateFormatter = new AjxDateFormat("yyyyMMdd");
+		var dateValue = AjxDateUtil._serverDateFormatter.format(new Date(date));
+		return dateValue;
```
miteshsavani810, line 1555:
```suggestion
		return AjxDateUtil._serverDateFormatter.format(new Date(date));
```

sumitTelus: Done

WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
```diff
@@ -1550,7 +1550,8 @@ function(inputs, conf, field) {
 		if (!date) {
 			return null;
 		}
-		return String(date.getTime() / 1000);
+		AjxDateUtil._serverDateFormatter = new AjxDateFormat("yyyyMMdd");
+		return AjxDateUtil._serverDateFormatter.format(new Date(date));
```
k-kato, line 1554: Can we use `AjxDateUtil.getServerDate` ?
https://github.com/Zimbra/zm-ajax/blob/develop/WebRoot/js/ajax/util/AjxDateUtil.js#L598

k-kato, line 1554: Can we use `AjxDateUtil.getServerDate` ?
https://github.com/Zimbra/zm-ajax/blob/develop/WebRoot/js/ajax/util/AjxDateUtil.js#L598
```suggestion
		return AjxDateUtil.getServerDate(new Date(date));
```

k-kato: ah, the function requires a string in `YYYYMMDDTHHMMSSZ` format...

k-kato: I saw a wrong result with the change.
1. created a filter specifying `Date before Dec/1/2023` in a condition.
2. a created filter has `Aug/17/1970` in a condition. zimbraMailSieveScript has also a wrong date.
```
$ zmprov ga test1 zimbraMailSieveScript
# name test1@dev.zimbra.com
zimbraMailSieveScript: require ["fileinto", "copy", "reject", "tag", "flag", "variables", "log", "enotify", "envelope", "body", "ereject", "reject", "relational", "comparator-i;ascii-numeric"];

# test
if anyof (date :before "19700823") {
    keep;
    stop;
}
```

sumitTelus: fixed the issue.

k-kato (changes requested):
See the comment https://github.com/Zimbra/zm-mailbox/pull/1542#pullrequestreview-1810712573
Server code needs to be updated first.
Please update UI code accordingly after that. (adding `tz` to `dateTest` , for example)

sumitTelus:
> See the comment [Zimbra/zm-mailbox#1542 (review)](https://github.com/Zimbra/zm-mailbox/pull/1542#pullrequestreview-1810712573) Server code needs to be updated first. Please update UI code accordingly after that. (adding `tz` to `dateTest` , for example)

@k-kato reverted the dataType of date in `dateTest` object back to Long from String.

k-kato:
Why do we need to update client code if timezone is calculated on server side?
The date is sent in long format in the current code, isn't it?

namita-uttarkar-telus:
> Why do we need to update client code if timezone is calculated on server side? The date is sent in long format in the current code, isn't it?

@k-kato 
The current implementation accepts a long value which is a combination of both date and time. We are dropping the time part and sending date alone in long format, since sending time part returns incorrect value.

WebRoot/js/zimbraMail/prefs/view/ZmFilterRuleDialog.js
```diff
@@ -1389,7 +1389,11 @@ function(ev) {
 		if (msg = this._checkCondition(c)) {
 			break;
 		} else {
-			rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
+			if (c.testType === ZmFilterRule.C_TEST_MAP[ZmFilterRule.C_DATE]) {
+				rule.addCondition(c.testType, c.comparator, Number(c.value), c.subjectMod, c.caseSensitive);
+			} else {
+				rule.addCondition(c.testType, c.comparator, c.value, c.subjectMod, c.caseSensitive);
+			}
```
k-kato, line 1396: Does a client need to send the date in long format as Number explicitly? 
Before the change, it is sent as String. Doesn't the server automatically convert it from a string it to a number?

k-kato (approved):

namita-uttarkar-telus:
> > Why do we need to update client code if timezone is calculated on server side? The date is sent in long format in the current code, isn't it?
> 
> @k-kato The current implementation accepts a long value which is a combination of both date and time. We are dropping the time part and sending date alone in long format, since sending time part returns incorrect value.

@k-kato We have dropped this condition and sending timeStamp in string format only.

PR #864, opened 2024-04-02
ZCS-14513: added error handling when disabling 2FA fails

k-kato:
**Issue:**
No error message is shown when disabling a 2FA method (app or email) fails.
For example, when a single 2FA method is enabled, zimbraFeatureTwoFactorAuthRequired is changed from FALSE to TRUE by administrator after the user logs in to web client and the user tries to disable the 2FA method, it fails because 2FA is mandatory. An error is returned from a server, but UI does not show any message.

**Changes:**
* updated `ZmTwoFactorSetupDialog.disableTwoFactorAuthCallback` to show an error message.
   * Note: it is called via `ZmTwoFactorSetupDialog.disableTwoFactorAuth` and `ZmAccountsPage.prototype._recoveryResetRequest`. For the latter, 2nd param `dialog` is null and 4th `result` is empty (undefined). `result` does not need to be checked.
* set a title in some dialogs. Previously `Information` or `Warning` was used.
* added `dialog.reset()` to remove a custom callback on buttons (OK/Cancel etc.).

**Related PR:**
* https://github.com/Zimbra/zm-ajax/pull/132

### DIFF
--- a/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
+++ b/WebRoot/js/zimbraMail/mail/view/prefs/ZmAccountsPage.js
@@ -307,6 +307,7 @@ ZmAccountsPage.prototype._handleTwoStepAuthLink =
 function(params, method, action) {
 	if (action === ZmAccountsPage.ACTION_DISABLE_TFA) {
 		var dialog = appCtxt.getYesNoMsgDialog();
+		dialog.reset();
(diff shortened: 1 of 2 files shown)


silentsakky (approved):

sitaram-mulik-synacor (approved):

HimanshuSojitraSynacor (approved):

lavegupta (approved):

k-kato:
squashed

PR #878, opened 2024-06-19
ZBUG-4053, ZBUG-4100: fixed to block delation of item while it is opened

k-kato:
**Issue:**
1) When a draft is being created and the draft is completely removed (shift + del key, or move it to Trash and run Empty Trash), SaveDraftRequest with message id fails as the file no longer exists.
2) When a message which has an attachment is being replied, the original message is removed, and click "Add attachments from original message" and do Save Draft, SaveDraftRequest fails with no such message error. 

Similar issues are observed in many places. These occur when an original message/attachment is removed during an action based on an existing item. For example,
* create a copy action on a message which has an attachment
* create appointment action on a message which has an attachment
* create task action on a message which has an attachment
* edit a contact
* forward appointment which has an attachment
* create a copy of appointment which has an attachment
* edit a task
* edit a document in briefcase

The data (db entry or blob file) must not be removed during the action. From viewpoint of Classic UI code, an original item must not be deleted permanently when it is internally used (as an object of JavaScript).

**Solution:**
Check if the deleting item is used in any tab or child window.
If it is used, the delete action is blocked. If not, the delete action can be executed.

**Code changes:**
* Add a process to update the status of controller. It has been implemented in most part, but was missing in some paths.
   * `inactive` is `false` when a tab/window is not shown. It is `true` when a tab/window is shown.
   * a controller is reused when a tab is closed and new tab is shown. A view object stored in the controller is also reused. Initialization process of controller and view are added.
* Add a logic to find existing tabs and windows to `ZmAppViewMgr.js`
* Add a logic to check if the deleting item(s) is used in any tab or window at hard deletion. (shift + del key, or remove from Trash)
* Add a logic to check if the deleting item(s) is used in any tab or window at Empty Folder and Delete Folder. It is searched recursively to check if the item is included in a sub folder of the target folder on Empty/Delete Folder.
* Regarding appointment, all delete action and Move To Trash is blocked when the appointment is used in a tab or window. This is because, processes of removing appointment(s) is complicated in Calendar app. There are many patterns that an appointment is moved to Trash or removed completely in a single action.
   * appointment with or without attendee
   * recurring appointment; series and instance
   * invited appointment
   * invited recurring appointment
   * and so on.
* Add a logic to check if the deleting attachment is used in any tab or window - actually checking if the message is used or not. If it is used, Remove (All) attachment action is blocked.
* When these permanent deletion is blocked, the message `One or more of the items you are trying to delete are open in a tab or window. Please close the item or items before permanently deleting.` is shown in a warning dialog.


silentsakky (approved):

WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
```diff
@@ -2181,6 +2191,33 @@ function(appt, mode) {
 	}
 };
 
+/**
+ * Check if deleting item(s) is opened
+ *
+ * @param {Array}   toDelete   list of items to delete
+ */
+ZmCalViewController.prototype.isDeletingItemOpen =
+function(toDelete) {
```
miteshsavani810, line 2200: Can we move the implementation of  `ZmCalViewController.prototype.isDeletingItemOpen` and `ZmBriefcaseController.prototype.isDeletingItemOpen` to the base class 'ZmController' ?

WebRoot/js/zimbraMail/briefcase/controller/ZmBriefcaseController.js
```diff
@@ -485,6 +489,21 @@ ZmBriefcaseController.prototype._doDelete = function(items, hardDelete) {
...
+ * @param {Array}   toDelete   list of items to delete
+ */
+ZmBriefcaseController.prototype.isDeletingItemOpen =
+function(toDelete) {
+	var deletingItemIds = [];
+	for (var i = 0; i < toDelete.length; i++) {
+		deletingItemIds.push(toDelete[i].id);
+	}
+	var appViewMgr = appCtxt.getAppViewMgr();
+	return appViewMgr && appViewMgr.isDeletingItemOpen(deletingItemIds);
```
miteshsavani810, line 504: I think we don't need to check `appViewMgr`. Can we directly call `isDeletingItemOpen`?.  Same for other places as well where we are checking the `appViewMgr`.

```suggestion
	return appCtxt.getAppViewMgr().isDeletingItemOpen(deletingItemIds);
```

WebRoot/js/zimbraMail/share/model/ZmList.js
```diff
@@ -730,6 +737,49 @@ function(params) {
...
+		viewIdToIgnore = appCtxt.getCurrentView().getController().tabId;
+	}
+	var appViewMgr = appCtxt.getAppViewMgr();
+	return appViewMgr && appViewMgr.isDeletingItemOpen(deletingItemIds, null, viewIdToIgnore);
+};
+
+/**
+ * Show a warning message dialog for item hard deletion
+ */
+ZmList.prototype._popupBlockItemDeletionWarningDialog =
```
miteshsavani810, line 772: Not sure. 
I think this is the general dialog so Can we move it to the `ZmAppViewMgr.js` ?

WebRoot/js/zimbraMail/briefcase/controller/ZmBriefcaseController.js
k-kato: fixed. Thank you for pointing it out.

WebRoot/js/zimbraMail/share/model/ZmList.js
k-kato: Moved the function to ZmAppViewMgr.

WebRoot/js/zimbraMail/calendar/controller/ZmCalViewController.js
k-kato: We can move them to a base class like ZmBaseController. However, `if - else if` statement needs to be added to handle controller-specific or view-specific process.
I think it is better to handle such specific process(es) in each child controller.

miteshsavani810 (approved):

damini-vashishtha (approved):

k-kato:
squashed commits and rebased the branch to develop

PR #896, opened 2025-09-08
ZBUG-4620:Fix domain name retrieval issue from server side

sonali033:
Issue:
Login was failing when users entered only their username without a domain while the "password must change" policy was enforced.

Fix Implemented:
Updated the UI to call a backend method, passing the server name.
On the server side, the domain is retrieved using the provided serverName.
The backend constructs the complete username.

### DIFF
--- a/WebRoot/public/login.jsp
+++ b/WebRoot/public/login.jsp
@@ -733,8 +733,9 @@ if (application.getInitParameter("offlineMode") != null) {
                                             String zimbraPasswordAllowedPunctuationChars = null;
 
                                             if (userName != null) {
+                                                String serverName = request.getServerName();
                                                 AccountSelector as = new AccountSelector(AccountBy.name, userName);
-                                                Account acct = Provisioning.getInstance().get(as);
+                                                Account acct = Provisioning.getInstance().get(as, serverName);
 
                                                 zimbraPasswordMinLength = acct.getPasswordMinLength();
                                                 zimbraPasswordMinUpperCaseChars = acct.getPasswordMinUpperCaseChars();


WebRoot/public/login.jsp
```diff
@@ -733,6 +733,11 @@ if (application.getInitParameter("offlineMode") != null) {
                                             String zimbraPasswordAllowedPunctuationChars = null;
 
                                             if (userName != null) {
+                                                String serverName = (String) pageContext.getAttribute("serverName");
+                                                if (userName.indexOf('@') == -1 && serverName != null) {
```
silentsakky: instead of this approach, can we process all this at the backend side

```suggestion
                                                String serverName = (String) pageContext.getAttribute("serverName");
                                                AccountSelector as = new AccountSelector(AccountBy.name, userName, serverName);
```
So AccountSelector will accept third parameter as server name and that will be used to get proper account details instead of jsp doing this logic


This way we don't have to expose domain information on jsp side and it can be internally used

silentsakky: CC @rcyarrapothu Let me know if this is ok for you

sonali033: done

WebRoot/public/login.jsp
```diff
@@ -733,8 +733,9 @@ if (application.getInitParameter("offlineMode") != null) {
                                             String zimbraPasswordAllowedPunctuationChars = null;
 
                                             if (userName != null) {
+                                                String serverName = (String) pageContext.getAttribute("serverName");
                                                 AccountSelector as = new AccountSelector(AccountBy.name, userName);
-                                                Account acct = Provisioning.getInstance().get(as);
+                                                Account acct = Provisioning.getInstance().get(as,serverName);
```
silentsakky:
```suggestion
                                                Account acct = Provisioning.getInstance().get(as, serverName);
```

k-kato:
zimbraPasswordMustChange is detected in [account/Auth.java](https://github.com/Zimbra/zm-mailbox/blob/10.1.10/store/src/java/com/zimbra/cs/service/account/Auth.java).
https://github.com/Zimbra/zm-mailbox/blob/10.1.10/store/src/java/com/zimbra/cs/service/account/Auth.java#L492

The Auth.java identifies a domain of the entered username if it does not have `@domain`.
```
if (acctEl != null) {
    acctValuePassedIn = acctEl.getText();
    acctValue = acctValuePassedIn;
    acctByStr = acctEl.getAttribute(AccountConstants.A_BY, AccountBy.name.name());
    acctBy = AccountBy.fromString(acctByStr);
    if (acctBy == AccountBy.name) {
        Element virtualHostEl = request.getOptionalElement(AccountConstants.E_VIRTUAL_HOST);
        String virtualHost = virtualHostEl == null ? null : virtualHostEl.getText().toLowerCase();
        if (virtualHost != null && acctValue.indexOf('@') == -1) {
            Domain d = prov.get(Key.DomainBy.virtualHostname, virtualHost);
            if (d != null)
                acctValue = acctValue + "@" + d.getName();
        }
    }
    acct = prov.get(acctBy, acctValue);
}
```
The logic should be the same for a username at password change. Using different logic may cause another issue, or make it difficult to find a cause of the issue.

The above code is long. It is better to add a new tag (class) in [zm-taglib](https://github.com/Zimbra/zm-taglib) repository.
Ideally, the `<% %>` block should be moved to zm-taglib.
https://github.com/Zimbra/zm-web-client/blob/10.1.10/WebRoot/public/login.jsp#L722-L758

Reference: functions called at login
https://github.com/Zimbra/zm-web-client/blob/10.1.10/WebRoot/public/login.jsp#L137
https://github.com/Zimbra/zm-taglib/blob/10.1.4/src/java/com/zimbra/cs/taglib/tag/LoginTag.java#L225
https://github.com/Zimbra/zm-mailbox/blob/10.1.10/client/src/java/com/zimbra/client/ZMailbox.java#L684
https://github.com/Zimbra/zm-mailbox/blob/10.1.10/client/src/java/com/zimbra/client/ZMailbox.java#L768

k-kato:
We discussed it in a different channel. The approach looks good to me.

As I commented in https://github.com/Zimbra/zm-mailbox/pull/1756, if there is a case that any exception is thrown by backend, frontend should handle it so that not an error page but a right error message can be shown.

> Ideally, the <% %> block should be moved to zm-taglib.
> https://github.com/Zimbra/zm-web-client/blob/10.1.10/WebRoot/public/login.jsp#L722-L758

It can be addressed in a separate ticket as refactoring.

k-kato:
Please also update Description of the PR. The fix is related to a flow of password change at login.

WebRoot/public/login.jsp
sonali033: done

sonali033:
> Please also update Description of the PR. The fix is related to a flow of password change at login.

Updated.

sonali033:
> It can be addressed in a separate ticket as refactoring.

I have created new ticket for code refactoring :
https://synacor.atlassian.net/browse/ZBUG-5122

k-kato (approved):

rcyarrapothu (approved):

silentsakky (approved):

silentsakky (approved):

sakshirawat0247 (approved):

PR #905, opened 2026-03-19
ZCS-18792: added QR code for 2FA setup

k-kato:
Add a function to show a QR code at 2FA setup.

**Fixes:**
* updated js code and css to show a QR code with a secret key for authenticator app during 2FA setup. It supports (A) 2FA setup in Preferences and (B) 2FA setup at login when it is mandatory.
* modified a message label
* added `twoFactorAuthQRCodeIssuer` for issuer of a label in topauth uri. A default value is "Zimbra".
* updated build.xml for new QRCode package

**Related PR:**
* https://github.com/Zimbra/zm-ajax/pull/139


WebRoot/css/zm.css
```diff
@@ -6057,6 +6057,18 @@ h2.ZmTwoFactorCustomTitleLoginPage {
...
 
+.ZmTwoFactorSetupContainer.ZmTwoFactorSetupAppKeyContainer {
+	margin: 0 10px;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.ZmTwoFactorSetupContainer.ZmTwoFactorSetupAppKeyContainer canvas.qrcode {
+	height: 135px !important;
```
silentsakky, line 6068: possible to remove !important statements from here?

WebRoot/css/zm.css
```diff
@@ -6075,6 +6087,15 @@ h2.ZmTwoFactorCustomTitleLoginPage {
 	color: @gray-darker@;
 }
 
+.ZmTwoFactorCustomSetupContainerLoginPage.ZmTwoFactorCustomSetupAppKeyContainer div.qrcode_wrapper {
+	display: inline-block;
+}
+
+.ZmTwoFactorCustomSetupContainerLoginPage.ZmTwoFactorCustomSetupAppKeyContainer canvas.qrcode {
+	height: 180px !important;
```
silentsakky, line 6095: same as above

silentsakky (approved):

WebRoot/css/zm.css
k-kato: `!important` is necessary to make it priority than element's style.
`QRCode.toCanvas` fucntion draws a QR code in a canvas element. It adds styles of a fixed value of height and width to the element. As far as I checked, `height:180px;` and `width:180px;` are set.
It may be possible to pass a width/height as a parameter for `QRCode.toCanvas`, but it is more flexible to use a css than hard-coded value.

WebRoot/css/zm.css
k-kato: same reason as the above

sakshirawat0247 (approved):

miteshsavani810 (approved):

PR #912, opened 2026-06-29
ZBUG-4269: Add missing splashScreenCopyright key for pt_BR locale

miteshsavani810:
pt_BR was falling back to the European Portuguese (pt) locale which uses 'registada' instead of the Brazilian Portuguese 'registrada'. Added the key explicitly to ZmMsg_pt_BR.properties and ZhMsg_pt_BR.properties with proper \uXXXX escaping for non-ASCII characters.

### DIFF
--- a/WebRoot/messages/ZhMsg_pt_BR.properties
+++ b/WebRoot/messages/ZhMsg_pt_BR.properties
@@ -1133,6 +1133,7 @@ spellcheckSuggestionsLabel = Sugest\u00f5es:
 
 splashScreenAppName = Cliente web
 splashScreenMobileAppName = Cliente web de celular
+splashScreenCopyright = Copyright \u00a9 2005-2026 Synacor, Inc. Todos os direitos reservados. "Zimbra" \u00e9 uma marca comercial registrada da Synacor, Inc.
 start = Inicial
 startLabel = Inicial:
 startDate = Data inicial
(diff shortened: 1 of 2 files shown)


k-kato (approved):

WebRoot/messages/ZhMsg_pt_BR.properties
```diff
@@ -1133,6 +1133,7 @@ spellcheckSuggestionsLabel = Sugest\u00f5es:
 
 splashScreenAppName = Cliente web
 splashScreenMobileAppName = Cliente web de celular
+splashScreenCopyright = Copyright \u00a9 2005-2026 Synacor, Inc. Todos os direitos reservados. "Zimbra" \u00e9 uma marca comercial registrada da Synacor, Inc.
```
silentsakky, line 1136: Not sure if this is in use somewhere

miteshsavani810: I can see only one place it has been used https://github.com/Zimbra/zm-web-client/blob/37e6fc50d05c3a229b4ecabe91eb55baa4591147/WebRoot/public/extuserprov.jsp#L131

This has been considered for Mobile version. We can remove this translation as mobile version support is deprecated.

silentsakky: Ok got it, external user provisioning is a supported flow. So let's continue with the changes.

silentsakky (approved):

PR #919, opened 2026-07-29
PREAPPS-9613: Added Philippines (Filipino/Tagalog) language support for Modern UI

miteshsavani810:


silentsakky:
Are we going to hide this option in Classic UI?

lavegupta (approved):

k-kato (approved):

miteshsavani810:
> Are we going to hide this option in Classic UI?

If we hide it then it might create a problem when Filipino user switch to Classic UI and check the Preference > Language field. Currently, For Filipino user, Classic UI show all UI element in english.

silentsakky (approved):

